PR #17, opened 2023-10-27
adds queries and charts

oaksprout:
Upgrades to pull data directly on chain

image


DavidMinarsch:
Current year should be 2 (I think it counts from 0).

DavidMinarsch:
Or we keep it year "1" but add an asterisk that it indexes from 0

oaksprout:
> Or we keep it year "1" but add an asterisk that it indexes from 0

I changed both the number and chart to index from 0, so that it's consistent with what the contracts return.

PR #44, opened 2024-03-20
fix: blogs bug fixes and improvements

mohandast52:
- These blogs are not loading in production, but should work now.
  - https://olas.network/blog/autonolas-community-update-002
  - https://olas.network/blog/cutting-through-the-hype

- Images should load in all the blogs

### DIFF
--- a/pages/blog/[id].jsx
+++ b/pages/blog/[id].jsx
@@ -35,7 +35,7 @@ const BlogItem = () => {
     <PageWrapper>
       <Meta pageTitle={title} siteImageUrl={imageUrl} />
       <div className="max-w-3xl mx-auto p-4">
-        {!imageError && (
+        {!imageError && imagePath && (
           <Image
             src={imageUrl}
             width={headerImage.data[0].attributes.formats.large.width}


oaksprout (approved):

styles/globals.js
```diff
@@ -24,6 +24,10 @@ export const markdownComponents = {
   h4: ({ node, ...props }) => <h4 className="text-lg font-medium mb-1" {...props} />,
   h5: ({ node, ...props }) => <h5 className="text-md font-medium" {...props} />,
   h6: ({ node, ...props }) => <h6 className="text-sm font-medium" {...props} />,
-  pre: ({ node, ...props }) => <pre className="p-4 bg-gray-800 text-white rounded-md overflow-auto" {...props} />,
+  pre: ({ node, ...props }) => <pre className="p-4 bg-gray-800 border rounded-md overflow-auto" {...props} />,
```
truemiller, line 27: I'm guessing some tailwind classes are being purged in production. We're using `bg-gray-800` which should create a dark grey background. 

Perhaps because the content is generated on the client and not used elsewhere we don't have access to this class?

Would be good to remove this in case it is resolved in the future and we end up with black text on a dark grey background.

truemiller (approved):

Tanya-atatakai (approved):

PR #63, opened 2024-04-23
Update dune data and use cases with new design

Tanya-atatakai:


tailwind.config.js
```diff
@@ -51,6 +54,7 @@ module.exports = {
       fontFamily: {
         body: ['Inter', 'system-ui', 'sans-serif'],
         heading: ['Manrope', 'sans-serif'],
+        sans: ['Inter', ...defaultTheme.fontFamily.sans],
```
Tanya-atatakai, line 57: please note this font style applied to the whole site as default, let me know if we need to have different fonts

components/SectionHeading.jsx
```diff
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 const SectionHeading = ({
   children, size, color, spacing, display, weight, other,
 }) => (
-  <h2 className={`${size} tracking-tight lg:text-heading ${spacing} ${color} ${display} ${weight} ${other}`}>
+  <h2 className={`tracking-tight md:text-heading ${size} ${spacing} ${color} ${display} ${weight} ${other}`}>
```
Tanya-atatakai, line 7: there's a global problem here that we use `md:` with `md:text-heading` (which has font size, line height, width, etc.) - we can't override such styles if we pass just e.g. `text-bold` to the `weight` - we need to pass it as `md:text-bold`
as it impacts other pages where SectionHeading  is used I'd prefer to fix it properly a bit later

oaksprout: Makes sense

oaksprout (changes requested):
Have added some comments via Vercel

truemiller (dismissed):

components/Layout/SectionWrapper.jsx
```diff
@@ -19,7 +23,18 @@ const SectionWrapper = ({
...
+    <section
+      className={`${customClasses} ${backgroundClasses} scroll-mt-[100px]`}
+      id={id}
+      style={
+        backgroundImage
+          ? {
+            background: `url(${backgroundImage}) center`,
+            backgroundSize: 'cover',
+          }
+          : undefined
```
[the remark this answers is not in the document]
Tanya-atatakai: I think better to use undefined here (like we didn't pass any properties at all), wdyt
it's not recommended to use inline styles in react components, I'd like to avoid passing empty object

tailwind.config.js
oaksprout: That's desirable yep

oaksprout:
Just one remaining issue on Vercel with the image resolution @Tanya-atatakai

oaksprout (approved):

truemiller (approved):

PR #85, opened 2024-07-12
Token page updates

Tanya-atatakai:
1) Supply sections now goes after hero, token details is the last one
2) Changed emissions per year bar chart title and description
3) Added supply chart
4) Updated Olas utility section wit new text and image
5) Also moved the Emissions per year chart to a separate file for better readability

(I also removed the old homepage folder, which I had kept previously because it had a component imported on the token page)

image 

image


components/OlasTokenPage/SupplyPieChart.jsx
```diff
@@ -0,0 +1,200 @@
...
+  'Valory (core contributor)',
+  'Circulating supply',
+];
+
+const ADDRESSES = [
+  veOlasAddress,
+  daoAddress,
+  buOlasAddress,
+  valoryAddress,
+];
```
truemiller, line 32: should merge these arrays into an array of objects so it's clear they're connected when they're matched by index later

components/OlasTokenPage/SupplyPieChart.jsx
```diff
@@ -0,0 +1,200 @@
...
+const COLORS = {
+  'bg-purple-500': '#A755F7',
+  'bg-pink-500': '#E964C4',
+  'bg-orange-400': '#FFB246',
+  'bg-green-400': '#3FE681',
+  'bg-cyan-500 ': '#09B4D7',
+};
```
truemiller, line 39: worthwhile making these a single array of objects as they're connected/used together and mapped over by index later.
i.e. 
`[{address: "", label: "", color:"" },...]`

components/OlasTokenPage/index.jsx
```diff
@@ -119,108 +101,31 @@ const Supply = () => {
...
-                      },
-                    }}
-                  />
-                )}
-              </div>
-
-              <p className="mb-4">
-                A maximum of 1bn OLAS tokens can be minted in the
-                protocol&apos;s first 10 years.
+                What are the maximum amount of OLAS that can be minted by the
```
truemiller, line 119:
```suggestion
                What is the maximum amount of OLAS that can be minted by the
```

truemiller (approved):

components/OlasTokenPage/SupplyPieChart.jsx
```diff
@@ -0,0 +1,200 @@
...
+                  ],
+                }}
+              />
+            )}
+          </div>
+
+          <div className="flex flex-col gap-2 self-center w-full">
+            {data.map((item, index) => (
+              <LegendItem
+                key={index}
```
truemiller, line 186: should avoid `key={index}` and use unique ids specific to the data.
may need to change the data once it's fetched, and for example, make it an object with identifier

shouldn't cause any issues though as it's not being filtered or sorted

components/OlasTokenPage/SupplyPieChart.jsx
Tanya-atatakai: great idea, updated

components/OlasTokenPage/EmissionScheduleChart.jsx
```diff
@@ -0,0 +1,144 @@
...
+        A maximum of 1bn OLAS tokens can be minted in the protocol&apos;s
+        first 10 years.
+      </p>
+      <p className="mb-4">
+        After year 10, an additional 2% can be minted each year. This 2%
+        inflation rate can be reduced by the DAO.
+      </p>
+      <div className="mb-4">
+        <Verify url="https://etherscan.io/address/0xc096362fa6f4A4B1a9ea68b1043416f3381ce300#readProxyContract#F19" />
+      </div>
```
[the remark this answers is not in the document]
Tanya-atatakai: can you please elaborate, what exactly?

components/OlasTokenPage/SupplyPieChart.jsx
```diff
@@ -0,0 +1,200 @@
...
+}
+
+const SupplyPieChart = () => {
+  const [data, setData] = useState([]);
+  const [totalSupply, setTotalSupply] = useState();
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    const fetchData = async () => {
+      const web3 = new Web3(new Web3.providers.HttpProvider(providerUrl));
```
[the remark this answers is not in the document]
Tanya-atatakai: I agree, that was annoying using this. But also think, let's maybe discuss that properly. It'd be a mess with all different approaches + we ideally need to switch to ts first to avoid all the problems we had with updating packages and etc.

components/OlasTokenPage/SupplyPieChart.jsx
```diff
@@ -0,0 +1,200 @@
...
+
+      const olasContract = new web3.eth.Contract(olasAbi, olasAddress);
```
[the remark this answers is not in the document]
Tanya-atatakai: moved to a separate file, please check if that suits

oaksprout (approved):

components/OlasTokenPage/SupplyPieChart.jsx
Tanya-atatakai: updated

truemiller (approved):

PR #99, opened 2024-08-07
fix: update shorts link on main page

Tanya-atatakai:


data/useCases.json
```diff
@@ -49,8 +49,7 @@
         "title": "Shorts.wtf",
         "description": "Entire short films, generated by AI agents.",
         "image": "/images/services/shorts.svg",
-        "link": "https://shorts.wtf",
-        "isExternal": true
+        "link": "/shorts"
```
DavidMinarsch, line 52: it's meant to be "services/shorts"

PR #128, opened 2024-09-05
style: updates to babydegen page

atepem:
## Proposed changes

Reformatted /babydegen page style in line with the new Figma wireframe. Changed headings, text, hero buttons, further resources cards and some of the layout. 

## Screenshots/Recordings
before:
https://github.com/user-attachments/assets/a6ddcb96-7035-4833-9610-05452408b801

after:
https://github.com/user-attachments/assets/67530858-9b88-4c47-9797-5c666b6aa125

### DIFF
--- a/pages/services/babydegen.jsx
+++ b/pages/services/babydegen.jsx
@@ -39,7 +39,7 @@ const HeaderLeftContent = () => (
         className="grow lg:w-auto "
       >
         <a href={OPERATE_AGENTS_URL}>
-          Explore available agents
+          Explore other agents today
         </a>
       </Button>
     </div>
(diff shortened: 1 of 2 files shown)


pages/services/babydegen.jsx
```diff
@@ -92,25 +107,28 @@ const WhatIsBabyDegen = () => (
 );
 
 const MeetTheOptimusAgent = () => (
-  <SectionWrapper customClasses="border-y border-b-0">
+  <SectionWrapper
+    customClasses="border-y border-b-0 border-[#FFE0E6]"
+    backgroundType="LINEAR_GRADIENT"
```
Tanya-atatakai, line 112: I think, there will be only one place where we'll have the red gradient, can we just use those class names in `customClasses`, instead of having it as a separate `backgroundType`?

atepem: This makes sense! I've changed changed that in the latest push now :)

Tanya-atatakai (approved):

Tanya-atatakai (approved):

oaksprout (approved):

PR #134, opened 2024-09-12
fix: update daa link, cache revalidation setting, role image

Tanya-atatakai:
## Proposed changes

1) Updated get involved image to the latest one
2) Updated the logic and link for daa metric
3) Added env variable for cache revalidation so we can change it in Vercel (set to 7200 sec = 2h)

## Things to keep in mind


components/HomepageSection/Activity.jsx
```diff
@@ -85,7 +85,7 @@ export const Activity = ({
           {dailyActiveAgents ? (
             <ExternalLink
               className="font-extrabold text-6xl"
-              href={`${DUNE_QUERY_URL}/${DAILY_ACTIVE_AGENTS_DUNE_QUERY_ID}`}
+              href={`${DUNE_QUERY_URL}/${DAILY_ACTIVE_AGENTS_DUNE_QUERY_ID}/6809279`}
```
DavidMinarsch, line 88: what's the hard coded value doing here?

DavidMinarsch (approved):

components/HomepageSection/Activity.jsx
Tanya-atatakai: @DavidMinarsch so that the tab with the average metric is shown. without it, the first tab will be open, which is the table

PR #135, opened 2024-09-12
feat: add use cases page

atepem:
## Proposed changes

Added use cases page and button link on homepage according to Figma wireframe. Also scroll-smooth added for smoother transitions from hero button to specified section.

## Screenshots/Recordings

Screenshot 2024-09-12 152931


components/HomepageSection/UseCases.jsx
```diff
@@ -105,6 +105,18 @@ const UseCases = () => (
...
+    </div>
+
+    <div>
+      <Button
+        variant="outline"
+        size="xl"
+        asChild
+        className="mt-12 max-sm:mt-3 col-start-2 mx-auto"
+      >
+        <Link href="/explore#use-cases">See more use cases</Link>
```
Tanya-atatakai, line 118: @atepem can you please confirm with Roman that it should open `/explore#use-cases` and not just `/explore`?

components/HomepageSection/UseCases.jsx
```diff
@@ -126,9 +138,7 @@ const UseCases = () => (
           The impact of autonomous agent economies is shaping applications and
           products in Web3 and far beyond.
         </p>
-        <Button variant="outline" size="xl" asChild>
-          <Link href="/explore#use-cases">See more use cases</Link>
-        </Button>
+
```
Tanya-atatakai, line 141: can you please remove this entire commented section? it previously contained the hexagon image but it's now moved to the explore page hero section

Tanya-atatakai (commented):
great work, lgtm!
can you please clarify one comment with Roman and also ask him if we need to uncomment some links we hid while we didn't have the explore page? it was in the menu and footer, you can find in the code by `link: '/explore`
if it's not needed, would be great if you removed the code

components/HomepageSection/UseCases.jsx
atepem: Roman's said that he'd rather it go to /explore instead of the specific section so I can change that now :)

common-util/constants.js
```diff
@@ -5,7 +5,7 @@ export const LAUNCH_CONTACT_URL = 'https://t.me/pahlmeyer';
 
 export const MENU_DATA = [
   { link: '/learn', text: 'Learn' },
-  // { link: '/explore', text: 'Explore' },
+  { link: '/explore', text: 'Explore' },
```
DavidMinarsch, line 8: No I don't want that back in the menu. This page is not that important to give it a menu bar section. If we want we can add it to "Resources" submenu

components/Layout/Footer.jsx
```diff
@@ -39,11 +39,9 @@ const LEARN_LINKS = [
   { title: 'Olas Stack', link: '/stack' },
 ];
 
-// const EXPLORE_LINKS = [
-//   { title: 'Use Cases', link: '/explore#use-cases' },
-//   { title: 'Chains', link: '/explore#chains' },
-//   { title: 'Dashboards', link: '/explore#dashboards' },
-// ];
+const EXPLORE_LINKS = [
+  { title: 'Use Cases', link: '/explore#use-cases' },
```
DavidMinarsch, line 43: If we call it "Use Cases" here, then in the submenu on the top it should have the same name

DavidMinarsch (changes requested):

common-util/constants.js
atepem: Okay, I can change that 👍 Is it alright to keep the footer link or should that be removed as well? 
image

atepem: I've removed Explore and put it under Resources instead

components/Layout/Footer.jsx
atepem: Yes it's been added to the Resources submenu and footer now

Tanya-atatakai (approved):

DavidMinarsch (approved):

PR #137, opened 2024-09-17
feat: migrate daa metric to flipside

Tanya-atatakai:
## Things to keep in mind


components/HomepageSection/Activity.jsx
```diff
@@ -1,38 +1,67 @@
...
+  };
+};
+
+const usePersistentSWR = (key, fetcher) => useSWR(key, fetcher, {
+  revalidateIfStale: false,
+  revalidateOnFocus: false,
+});
+
+export const Activity = () => {
+  const { data: metrics } = usePersistentSWR('activityMetrics', fetchMetrics);
```
Tanya-atatakai, line 48: use swr here so the data is cached and not re-requested when we visit another page and go back

truemiller (approved):

DavidMinarsch (approved):

PR #139, opened 2024-09-19
feat: add links for sections on homepage

Tanya-atatakai:
## Proposed changes

supports the following links which scroll the page to the corresponding sections:

- olas.network#usecases
- olas.network#tech
- olas.network#ecosystem
- olas.network#growth

## Things to keep in mind

### DIFF
--- a/components/HomepageSection/PropelledBy.jsx
+++ b/components/HomepageSection/PropelledBy.jsx
@@ -107,7 +107,7 @@ const MoreFriends = () => (
 );
 
 export const PropelledBy = () => (
-  <SectionWrapper>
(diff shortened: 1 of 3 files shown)


components/HomepageSection/Activity.jsx
```diff
@@ -88,7 +88,7 @@ export const Activity = () => {
   return (
     <SectionWrapper
       customClasses="text-center py-16 px-4 border-b"
-      id="activity"
```
DavidMinarsch, line 91: ok we can leave it as it was. I was not aware this had one

DavidMinarsch (approved):

PR #151, opened 2024-10-23
feat: mech metrics

atepem:
## Proposed changes

Added metrics for mech page: predict, contribute, governatooor and other.

## Screenshots/Recordings

Before:
image
After:
image

### DIFF
--- a/common-util/constants.js
+++ b/common-util/constants.js
@@ -21,10 +21,13 @@ export const CORE_TECHNICAL_DOCUMENT =
 export const FLIPSIDE_URL =
   'https://flipsidecrypto.xyz/flipsideteam/olas-key-activity-metrics-pnPjda';
 export const DUNE_QUERY_URL = 'https://dune.com/queries';
-export const PREDICTION_DAA_QUERY_ID = '4165113';
+export const PREDICTION_DAA_QUERY_ID = '4185253';
 export const PREDICTION_TXS_BY_AGENT_TYPE_QUERY_ID = '4161414';
 export const PREDICTION_ECONOMY_DASHBOARD_URL =
   'https://dune.com/adrian0x/the-olas-predict-agent-economy';
+export const MECH_RQS_QUERY_ID = '4185253';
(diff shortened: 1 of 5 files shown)


components/MechPage/MechAgentMetrics.jsx
```diff
@@ -1,24 +1,159 @@
...
+  return {
+    totalRqs: result.totalRqs,
+    predictRqs: result.predictRqs,
+    contributeRqs: result.contributeRqs,
+    governatooorRqs: result.governatooorRqs,
+    otherRqs: result.otherRqs,
+  };
+};
+
+const usePersistentSWR = (key, fetcher) =>
```
truemiller, line 21: can be made reusable, but minor

truemiller (approved):

components/MechPage/MechAgentMetrics.jsx
Tanya-atatakai: i'd rather switch to react tanstack like in mono repo

common-util/constants.js
```diff
@@ -21,10 +21,13 @@ export const CORE_TECHNICAL_DOCUMENT =
 export const FLIPSIDE_URL =
   'https://flipsidecrypto.xyz/flipsideteam/olas-key-activity-metrics-pnPjda';
 export const DUNE_QUERY_URL = 'https://dune.com/queries';
-export const PREDICTION_DAA_QUERY_ID = '4165113';
+export const PREDICTION_DAA_QUERY_ID = '4185253';
```
Tanya-atatakai, line 24: was this changed accidentally? the value looks the same as mech query's, but I think should be the old one, the metric is also broken in Vercel

Tanya-atatakai (changes requested):

common-util/constants.js
atepem: Yes, I think it was changed on accident. I'll change this in a new PR real quick

PR #153, opened 2024-10-24
feat: Actual Emissions Sched chart added to olas-token page

atepem:
## Proposed changes

Added Actual Emissions Schedule chart that shows sum of actual emissions and sum of available. Changed placement of "Dive into current distribution details" part.

## Screenshots/Recordings

image
image

### DIFF
--- a/components/OlasTokenPage/ActualEmissionsChart.jsx
+++ b/components/OlasTokenPage/ActualEmissionsChart.jsx
@@ -1,0 +1,90 @@
+import {
+  Chart,
+  Filler,
+  LineElement,
+  LinearScale,
+  PointElement,
+  Tooltip,
+} from 'chart.js';
+import {
+  EMISSIONS_CHART_COLORS,
+  getEmissionsChartOptions,
+} from 'common-util/charts';
+import PropTypes from 'prop-types';
+import { Line } from 'react-chartjs-2';
+import { LegendItem } from './LegendItem';
+import { emissionType } from './types';
+
+Chart.register(LineElement, LinearScale, PointElement, Filler, Tooltip);
+
+export const ActualEmissionsChart = ({ emissions, loading }) => {
+  const maxAvailableEmissions = emissions.map(
+    (item) =>
+      parseFloat(item.availableDevIncentives || 0) +
+      parseFloat(item.availableStakingIncentives || 0) +
+      parseFloat(item.effectiveBond || 0),
+  );
+  const actualEmissions = emissions.map(
+    (item) =>
+      parseFloat(item.devIncentivesTotalTopUp || 0) +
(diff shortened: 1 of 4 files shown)


components/OlasTokenPage/ActualEmissionsChart.jsx
```diff
@@ -0,0 +1,90 @@
...
+                    order: 2,
+                    pointBackgroundColor: EMISSIONS_CHART_COLORS.actual.line,
+                    borderColor: EMISSIONS_CHART_COLORS.actual.line,
+                  },
+                ],
+              }}
+              options={getEmissionsChartOptions([
+                ...maxAvailableEmissions,
+                ...maxAvailableEmissions,
+                10 ** 18, // Add this temporarily while we don't have data on staking
```
Tanya-atatakai, line 77: this can be removed since we have all the data, it was used only for one epoch when we didn't have data on the staking chart but wanted to show a higher line

components/OlasTokenPage/ActualEmissionsChart.jsx
```diff
@@ -0,0 +1,90 @@
...
+import { Line } from 'react-chartjs-2';
+import { LegendItem } from './LegendItem';
+import { emissionType } from './types';
+
+Chart.register(LineElement, LinearScale, PointElement, Filler, Tooltip);
+
+export const ActualEmissionsChart = ({ emissions, loading }) => {
+  const maxAvailableEmissions = emissions.map(
+    (item) =>
+      parseFloat(item.availableDevIncentives || 0) +
```
Tanya-atatakai, line 23: better to avoid parseFloat and use BigInt to maintain precision, e.g. `BigInt(item.availableDevIncentives || 0) + ...`

atepem: This returns a type error as the Line component does not take bigint values

components/OlasTokenPage/index.jsx
```diff
@@ -150,6 +151,21 @@ const Supply = () => {
             />
           </div>
 
+          <div className="border rounded-lg">
+            <div className="p-4 border-b">
+              <h2 className="text-xl mb-2 font-bold">
+                Actual Emissions Schedule
+              </h2>
+              <p className="text-slate-500">
+                What are the actual amount of OLAS minted by the protocol per
```
truemiller, line 160:
```suggestion
                What is the actual amount of OLAS minted by the protocol per
```

components/OlasTokenPage/ActualEmissionsChart.jsx
atepem: Done now :)

truemiller (approved):
nice work!

DavidMinarsch (approved):

PR #194, opened 2025-01-20
style: fixing text inconsistencies

atepem:
## Proposed changes

Changes to text for better tone and consistent formatting on following pages:
* AU
* Bond
* Q Updates
* Explore
* FAQ
* Stack
* Protocol
* Learn
Also fixed hydration error on build page academy metrics table.

### DIFF
--- a/components/BondsPage/HowBondingWorks.jsx
+++ b/components/BondsPage/HowBondingWorks.jsx
@@ -1,14 +1,15 @@
-import { Fragment } from 'react';
 import Link from 'next/link';
+import { Fragment } from 'react';
 
-import { BONDING_PROGRAMS_URL } from 'common-util/constants';
 import {
-  SECTION_BOX_CLASS,
   SCREEN_WIDTH_LG,
+  SECTION_BOX_CLASS,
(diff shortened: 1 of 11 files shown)


components/BondsPage/HowBondingWorks.jsx
```diff
@@ -51,7 +63,7 @@ export const HowBondingWorks = () => (
       {steps.map(({ title, description }) => (
         <Fragment key={title}>
           <h3 className="text-xl font-semibold mt-4">{title}</h3>
-          <p>{description}</p>
+          <div>{description}</div>
```
Tanya-atatakai, line 66: doesn't it work with p tag?

Tanya-atatakai (approved):

components/BondsPage/HowBondingWorks.jsx
atepem: yes it does,, I think I just changed it bc if the description contains a div it makes a <p><div> hydration error 😓

DavidMinarsch (approved):

PR #201, opened 2025-01-29
style: added /academy page

atepem:
## Proposed changes

Migrated /academy page from valory site to olas.

## Screenshots/Recordings

image

## Things to keep in mind


DavidMinarsch (approved):

styles/globals.js
```diff
@@ -19,12 +19,7 @@ export const markdownComponents = {
     <a className="text-purple-800 hover:text-blue-800" {...props} />
   ),
   // Apply margin to paragraphs to create space between them
-  p: ({ node, ...props }) => (
-    <>
-      <p className="inline" {...props} />
-      <br />
```
Tanya-atatakai, line 25: don't we break something with this change? I remember you added it to fix posts or something?

Tanya-atatakai (approved):

styles/globals.js
atepem: Yes I remember this as well, but I've checked through the posts and they don't appear to be affected :?

PR #236, opened 2025-04-02
Style: added Podcast section to AU page

atepem:
## Proposed changes

Podcast section added to AU page. Currently takes from videos collection from CMS using useFetchVideos hook and then filters via titles containing "Ep." as a workaround until we can get a specific collection for Podcasts. 
Currently using empty state as no podcasts available yet.

## Screenshots/Recordings

When empty:
image

When filled:
image


common-util/useFetchApi.js
```diff
@@ -13,12 +13,23 @@ const useFetch = (subUrl, params) => {
   return { data, isLoading };
 };
 
-export const useFetchVideos = (limit = 1000) => {
-  const params = {
-    sort: ['date:desc'],
-    populate: '*',
-    'pagination[limit]': limit,
-  };
+export const useFetchVideos = (limit = 1000, isPodcast = false) => {
```
Tanya-atatakai, line 16: would be nice to make params an object, so that when in use it's easier to read this `useFetchVideos({ limit: LIMIT, isPodcats: true })` rather than this `useFetchVideos(LIMIT, true)`

common-util/useFetchApi.js
```diff
@@ -13,12 +13,23 @@ const useFetch = (subUrl, params) => {
...
+          title: {
+            $contains: 'Ep.',
+          },
+        },
+      }
+    : {
+        sort: ['date:desc'],
+        populate: '*',
+        'pagination[limit]': limit,
+      };
```
Tanya-atatakai, line 32: tiny and up to you, could be just below to safe some space + for readability
```suggestion
  const params = {
        sort: ['date:desc'],
        populate: '*',
        'pagination[limit]': limit,
        filters: isPodcast ? {
          title: {
            $contains: 'Ep.',
          },
        } : undefined,
      }
```

Tanya-atatakai (approved):

Tanya-atatakai (approved):

DavidMinarsch (approved):

PR #239, opened 2025-04-10
style: display fee flows on MM

atepem:
## Proposed changes

* Mech marketplace II: improve mech marketplace dashboard to show fee flow
  * Mech fees - make clear on site how they are collected; what level they are at (show don’t tell); how this connects to Olas burn

Added Sankey diagram to mech marketplace page.

### DIFF
--- a/components/MechPage/FeeMetrics.jsx
+++ b/components/MechPage/FeeMetrics.jsx
@@ -1,0 +1,150 @@
+import { SUB_HEADER_CLASS } from 'common-util/classes';
+import { MECH_ECONOMY_DASHBOARD_URL } from 'common-util/constants';
+import SectionWrapper from 'components/Layout/SectionWrapper';
+import Link from 'next/link';
+import { Chart } from 'react-google-charts';
+
+const formerData = [
+  {
+    id: 'total-fees',
+    label: (
+      <div className="flex flex-col gap-2 mb-3">
+        <span className="text-base font-semibold ">
+          Total Agent Fees Collected
+        </span>
+      </div>
+    ),
+    // value: metrics?.predictTxs?.toLocaleString(),
+    value: 38830,
+    source: MECH_ECONOMY_DASHBOARD_URL,
+    color: '#7a9cf7',
+  },
+  {
+    id: 'unclaimed',
+    label: (
+      <div className="flex flex-col gap-2 mb-3">
+        <span className="text-base font-semibold ">Unclaimed Fees</span>
+      </div>
+    ),
+    // value: metrics?.contributeTxs?.toLocaleString(),
+    value: 23298,
+    source: MECH_ECONOMY_DASHBOARD_URL,
+    color: '#d96a5e',
+  },
+  {
+    id: 'claimed',
+    label: (
+      <div className="flex flex-col gap-2 mb-3">
+        <span className="text-base font-semibold ">Claimed Fees</span>
+      </div>
+    ),
+    // value: metrics?.governatooorTxs?.toLocaleString(),
+    value: 15532,
+    source: MECH_ECONOMY_DASHBOARD_URL,
+    color: '#5fb178',
+  },
+  {
+    id: 'received',
+    label: (
+      <div className="flex flex-col gap-2 mb-3">
+        <span className="text-base font-semibold ">Fees Receieved</span>
+      </div>
+    ),
+    // value: metrics?.otherTxs?.toLocaleString(),
+    value: 15376.67,
+    source: MECH_ECONOMY_DASHBOARD_URL,
+    color: '#68bcce',
+  },
+  {
+    id: 'olas-burned',
+    label: (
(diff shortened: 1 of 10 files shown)


components/MechPage/FeeMetrics.jsx
```diff
@@ -0,0 +1,216 @@
...
+      formerData[2].value * 0.1,
+      `Claimed Fees → OLAS Burned | $${formerData[2].value * 0.1} (1%)`,
+    ],
+    [
+      'Claimed Fees',
+      'Fees Received',
+      formerData[2].value * 0.9,
+      `Claimed Fees → Fees Received | $${formerData[2].value * 0.9} (99%)`,
+    ],
+  ];
```
Tanya-atatakai, line 108: this is hard to review, would be nice to leave some comments on what it does/what it's for, or maybe move small pieces to functions for better readability

components/MechPage/FeeMetrics.jsx
```diff
@@ -0,0 +1,216 @@
...
+    handleResize();
+    window.addEventListener('resize', handleResize);
+    return () => window.removeEventListener('resize', handleResize);
+  }, []);
+
+  const chartSizes = () => {
+    if (windowWidth < 640) return { fontSize: 12, chartHeight: 200 };
+    if (windowWidth < 1024) return { fontSize: 16, chartHeight: 300 };
+    return { fontSize: 20, chartHeight: 400 };
+  };
```
Tanya-atatakai, line 123: would be nice to move it to a custom hook and call it similar to what we have in mono repo, e.g. useScreen
so that it's easier to write code in different apps, if the same functions are used

components/MechPage/FeeMetrics.jsx
```diff
@@ -0,0 +1,216 @@
...
+      {error ? (
+        <div className="text-center py-8">Error loading metrics</div>
+      ) : (
+        <>
+          <div className="w-full max-w-full overflow-x-auto mb-8">
+            <div className="min-w-[320px] max-w-7xl mx-auto w-full overflow-hidden">
+              <Chart
+                chartType="Sankey"
+                width="100%"
+                height={chartSizes().chartHeight}
```
Tanya-atatakai, line 181: can you please check if it works fine, I see the shart is 300px height but the content inside is smaller 
image

components/MechPage/FeeMetrics.jsx
```diff
@@ -0,0 +1,216 @@
...
+                  className={`text-start flex flex-col w-[280px] p-3 border-gray-300 h-full max-sm:w-full ${borderClassName}`}
+                  style={{ color: item.color }}
+                >
+                  <div className="flex flex-col gap-2 mb-3">
+                    <span className="text-base max-sm:text-sm flex flex-wrap text-black font-semibold">
+                      {item.label}
+                    </span>
+                  </div>
+                  <span className="block text-4xl whitespace-nowrap max-sm:text-xl font-extrabold mb-4 mt-auto">
+                    $ {item.value} ↗
```
Tanya-atatakai, line 206: shouldn't they be links? as we have "↗"

components/MechPage/FeeMetrics.jsx
```diff
@@ -0,0 +1,216 @@
...
+        </p>
+      </div>
+
+      {error ? (
+        <div className="text-center py-8">Error loading metrics</div>
+      ) : (
+        <>
+          <div className="w-full max-w-full overflow-x-auto mb-8">
+            <div className="min-w-[320px] max-w-7xl mx-auto w-full overflow-hidden">
+              <Chart
```
Tanya-atatakai, line 178: I see some issues in the chart:
1) can you check what is this, looks like some data is not displayed 
image

2) when I hover over these two, I see the same labels, is it expected?
image

components/MechPage/FeeMetrics.jsx
atepem: This does work, the issue is that since there are no claimed fees yet, the chart makes room for something that is 0 eventhough its one whole non-branch diagram rn. This is what it would look like if the chart had claimed fees:
image
If I adjust to only up to that height, the rest of these numbers would be cut off:
image

components/MechPage/FeeMetrics.jsx
atepem: Used Links -- ExternalLinks doesn't work as it overwrites the number colours which doesn't work as they correspond to the branch colours as well. Ty for catching!

components/MechPage/FeeMetrics.jsx
atepem: 1. I can't reproduce the first issue :/
2. I'd added the 2nd branch so that the gradient would stop where Iason asked for it to, but I think we can remove this and make this look smoother, with the drawback of the overall sankey diagram changing:
image
(with dummy data ^)
image
(without ^)

atepem:
Will address these comments in a future PR

DavidMinarsch (approved):

PR #245, opened 2025-04-25
chore: added csp security headers

atepem:
## Proposed changes

Added CSP headers to `next.config.js`


atepem:
Misunderstood the task -- can revisit this and fix issues when adding security headers for sites.

PR #247, opened 2025-05-01
feat: update emisisons chart

Tanya-atatakai:
## Proposed changes

updates all emissions charts so that they show cumulative data by epochs 

## Things to keep in mind

### DIFF
--- a/common-util/charts.js
+++ b/common-util/charts.js
@@ -1,7 +1,4 @@
-import {
-  formatChartNumber,
-  formatWeiNumber,
-} from 'common-util/numberFormatter';
+import { formatWeiNumber } from 'common-util/numberFormatter';
 
 export const EMISSIONS_CHART_COLORS = {
   available: {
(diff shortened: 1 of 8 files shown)


common-util/charts.js
```diff
@@ -75,43 +72,3 @@ export const getEmissionsChartOptions = (points) => ({
     },
   },
 });
-
-export const getEmissionsChartOptionsFromNumber = () => ({
```
Tanya-atatakai, line 79: I removed it and `formatChartNumber` as it seems it was introduced specifically for all emissions to fix some number formatting issue. used the existing one and fixes the formatting where data is prepared but @atepem please lmk if I miss something

atepem (dismissed):

common-util/charts.js
atepem: lgtm :) I think it's fine to leave those off

mariapiamo (approved):

PR #248, opened 2025-05-02
feat: timeline page

atepem:
## Proposed changes

Added new timeline page on /timeline.

## Screenshots/Recordings

image
image

## Things to keep in mind

- [X] I confirm I have updated the meta title and description for the page, if applicable.

### DIFF
--- a/components/TimelinePage/Content.jsx
+++ b/components/TimelinePage/Content.jsx
@@ -91,7 +91,7 @@ const Categories = ({ toggleFilters }) => (
 const Timeline = ({ filters, setFilters }) => {
   return (
     <SectionWrapper customClasses="pt-0 py-12">
-      <div className="max-w-[872px] mx-auto max-md:mx-4">
+      <div className="max-w-[872px] mx-auto max-lg:mx-4">
         <div className="justify-end flex flex-row place-items-center mb-10 gap-4">
           <div>{filters.length} filters applied.</div>
           <Button
@@ -108,7 +108,7 @@ const Timeline = ({ filters, setFilters }) => {
             className="flex flex-col md:flex-row gap-12 md:gap-[120px] w-full"
           >
             <h3
-              className={`${HEADER_LARGE_CLASS} sticky top-0 h-fit max-md:text-center`}
+              className={`${HEADER_LARGE_CLASS} md:sticky top-[120px] h-fit max-md:text-center`}
             >
               {item.year}
             </h3>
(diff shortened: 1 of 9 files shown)


components/TimelinePage/Content.jsx
```diff
@@ -0,0 +1,174 @@
...
+        <div className="justify-end flex flex-row place-items-center mb-10 gap-4">
+          <div>{filters.length} filters applied.</div>
+          <Button
+            variant="outline"
+            className="px-1"
+            onClick={() => setFilters([])}
+          >
+            Reset all
+          </Button>
+        </div>
```
Tanya-atatakai, line 104: I didn't find how we apply the filters, is this needed? if so, I suggest to hide both text and button when 0 filters are applied

Tanya-atatakai: now I see how it works, we should click the categories. I would say it's not intuitive :D but anyway, can we:
1) hide the text and button when 0 filters are applied
2) highlight chosen filters in categories so it's clearer they work as filters
3) display only quarters where something has been found after applying filters, and if nothing found - display a message?

components/TimelinePage/Content.jsx
```diff
@@ -0,0 +1,174 @@
...
+            </h3>
+            <div className="w-full flex flex-col">
+              {item.quarters.map((quarter, index) => (
+                <div key={`quarter ${index}`} className="mb-14">
+                  <div className="w-full justify-between bg-slate-100 rounded-lg flex flex-row place-items-center px-6 py-4 mb-4">
+                    <h4 className={`${TEXT_LARGE_CLASS} font-bold`}>
+                      {quarter.quarter}
+                    </h4>
+                    <p className="text-slate-600">
+                      {quarter.date || getQuarterLength(quarter.quarter)}
```
Tanya-atatakai, line 123: to simplify things, do you think we could just simply hardcode this text in timeline.json? e.g. have `date: "Jan - Mar"` or `"date": "September",` so that we won't need to have `quarterRanges` and `getQuarterLength`. wdyt @atepem @mohandast52

atepem: This works too! I added the "date" bit last second so i didn't realise I could also do it that way too

components/TimelinePage/Content.jsx
atepem: Ok, will check Roman's updated design

Tanya-atatakai (approved):

DavidMinarsch (commented):
Please link to it from "More Resources"

Tanya-atatakai (approved):

DavidMinarsch (approved):

PR #261, opened 2025-05-21
chore: added cachedResponse as fallback for APR metrics

atepem:
## Proposed changes

Last updated fallback for faster APR metrics loading.

### DIFF
--- a/components/BabydegenEconomyPage/OptimusAgentMetrics.jsx
+++ b/components/BabydegenEconomyPage/OptimusAgentMetrics.jsx
@@ -23,12 +23,20 @@ const fetchMetrics = async () => {
 };
 
 const formatNumber = (num) => {
+  if (num === null || num === undefined) return null;
   const numTo1dp = Number(num?.toFixed(1));
   return `${numTo1dp}%`;
 };
 
 export const OptimusAgentMetrics = () => {
-  const { data: metrics } = usePersistentSWR('OptimusMetrics', fetchMetrics);
+  const {
+    data: metrics,
+    isLoading,
+    error,
+  } = usePersistentSWR('OptimusMetrics', fetchMetrics, {
+    refreshInterval: 3600000, // Refresh every hour
+    revalidateOnFocus: true,
+  });
 
   const data = useMemo(
     () => [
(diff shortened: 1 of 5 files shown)


components/BabydegenEconomyPage/OptimusAgentMetrics.jsx
```diff
@@ -23,12 +23,16 @@ const fetchMetrics = async () => {
...
+  if (num === null || num === undefined) return null;
   const numTo1dp = Number(num?.toFixed(1));
   return `${numTo1dp}%`;
 };
 
 export const OptimusAgentMetrics = () => {
-  const { data: metrics } = usePersistentSWR('OptimusMetrics', fetchMetrics);
+  const { data: metrics } = usePersistentSWR('OptimusMetrics', fetchMetrics, {
+    refreshInterval: 3600000, // Refresh every hour
+    revalidateOnFocus: true,
```
Tanya-atatakai, line 34: I don't see this parameters are used in usePersistentSWR, will they work as expected?

pages/api/optimus-metrics.js
```diff
@@ -25,24 +28,95 @@ export default async function handler(req, res) {
...
+      },
+      timestamp: Date.now(),
+    };
+
+    // Save to cache file
+    try {
+      fs.writeFileSync(CACHE_FILE_PATH, JSON.stringify(data, null, 2));
+    } catch (writeError) {
+      console.error('Error writing to cache:', writeError);
+    }
```
Tanya-atatakai, line 60: let's use nextjs for caching instead of writing to file https://nextjs.org/docs/app/deep-dive/caching#time-based-revalidation

can even remove Cache-Control header from here, as afaik Vercel should resolve this automatically

Tanya-atatakai (changes requested):

pages/api/optimus-metrics.js
atepem: Oh thanks this was exactly what I was looking for! Can remove the majority of the unnecessary stuff now :D

pages/api/optimus-metrics.js
```diff
@@ -25,24 +28,95 @@ export default async function handler(req, res) {
     const totalTraces = traces.length;
 
     if (totalTraces < MIN_TOTAL_TRACES) {
-      return res.status(404).json({ error: 'Not enough data traces found' });
+      console.error('Not enough traces found:', totalTraces);
+      return null;
```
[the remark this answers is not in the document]
atepem: have changed to `return res.status(404).json({ error: 'Not enough data traces found' });`

PR #283, opened 2025-06-11
style: make 0.0 the default value unless fresh data is available

atepem:
## Proposed changes

Make the default number for babydegen APR metrics 0.0% unless fresh data is available through cache file or from fetching from HuggingFace.

Also fixed numbered verify buttons on tokenomics page under **Current Usage** -> **Per epoch distribution**.

### DIFF
--- a/components/BabydegenEconomyPage/OptimusAgentMetrics.jsx
+++ b/components/BabydegenEconomyPage/OptimusAgentMetrics.jsx
@@ -29,11 +29,10 @@ const formatNumber = (num) => {
 };
 
 export const OptimusAgentMetrics = () => {
-  const { data: metrics } = usePersistentSWR(
-    'OptimusMetrics',
-    fetchMetrics,
-    true,
-  );
+  const { data: metrics } = usePersistentSWR('OptimusMetrics', fetchMetrics, {
+    refreshInterval: 10000, // refresh every 10s
+    dedupingInterval: 5000,
+  });
 
   const data = useMemo(
     () => [
--- a/hooks/index.js
+++ b/hooks/index.js
@@ -1,11 +1,11 @@
 import { useEffect, useState } from 'react';
 import useSWR from 'swr';
 
-export const usePersistentSWR = (key, fetcher, refresh = false) =>
+export const usePersistentSWR = (key, fetcher, config) =>
   useSWR(key, fetcher, {
     revalidateIfStale: false,
     revalidateOnFocus: false,
-    ...(refresh && { refreshInterval: 30000, dedupingInterval: 5000 }),
+    ...config,
   });
 
 export const useWindowWidth = () => {


data/optimus-metrics-cache.json
```diff
@@ -3,5 +3,5 @@
     "latestAvgApr": -12.345452875000001,
     "latestETHApr": -8.108115914125001
```
Tanya-atatakai, line 4: with be more logical with the update. then once the page is visited by someone, the file is updated on Vercel side
```suggestion
    "latestAvgApr": 0.00,
    "latestETHApr": 0.00
```

components/OlasTokenPage/UsagePieChart.jsx
```diff
@@ -98,7 +98,7 @@ export const UsagePieChart = ({ epoch, split, loading }) => (
       <div className="flex flex-row gap-3 text-slate-400 mb-4">
         <p>Verify: </p>
         {verifyLinks.map((link, index) => (
-          <Verify key={`Verify ${index + 1}`} url={link} text={index} />
+          <Verify key={`Verify ${index}`} url={link} text={index + 1} />
```
Tanya-atatakai, line 101: thanks!

pages/api/optimus-metrics.js
```diff
@@ -86,11 +86,6 @@ export default async function handler(req, res) {
 
     if (isCacheExpired) {
       fetchLatestMetric().catch(console.error);
-
-      // Returns cached data, even expired
-      if (cachedData) {
-        return res.status(200).json(cachedData.data);
-      }
```
Tanya-atatakai, line 93: no need to remove it - we could still show the cached data, it'll be refetched anyway if stale, see this: https://valory-workspace.slack.com/archives/C046R84E056/p1749659521162749?thread_ts=1749482255.353949&cid=C046R84E056

so with this PR could just set 0.00 to the data/optimus-metrics-cache.json
sorry for confusion

components/BabydegenEconomyPage/OptimusAgentMetrics.jsx
```diff
@@ -29,7 +29,13 @@ const formatNumber = (num) => {
 };
 
 export const OptimusAgentMetrics = () => {
-  const { data: metrics } = usePersistentSWR('OptimusMetrics', fetchMetrics);
+  const containerRef = useRef(null);
+  const { data: metrics } = usePersistentSWR(
+    'OptimusMetrics',
+    fetchMetrics,
+    true,
+    containerRef,
```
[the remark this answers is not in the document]
atepem: For some reason the changes made to usePersistentSWR hook was not added in with this PR, but true is supposed to refer to whether a refresh was needed when updating the APR data. I think it had `refereshInterval` and `dedupingInterval` set to 30s and 5mins respectively -- do you think this is worth adding so that the numbers change on page without requiring refresh?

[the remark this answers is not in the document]
Tanya-atatakai: from what I see, we better extend our usePersistentSWR with a third param called "config" and pass it to 'useSWR' instead of having two extra parameters, meaning of which is hard to get just reading the function usage.
if you go to the definition of useSWR, you will see here the extra config option can be passed
image

the list of possible options you can see in `PublicConfiguration` here `node_modules/swr/_internal/dist/index.d.ts`
in our function there's already some default values for the config, so you can do this:
```
{
    revalidateIfStale: false,
    revalidateOnFocus: false,
    ...config
}
```

I don't understand what true means as well as the ref, it's not in the list of config options. however you can pass both `refreshInterval` and `dedupingInterval` if you want. if needed, please DM and we can sync on it more deeply

Tanya-atatakai (approved):

PR #290, opened 2025-06-17
style: added olas: powers ai section

atepem:
## Proposed changes

Added OLAS: Powers AI section and mobile version.

## Screenshots/Recordings

image
image

Mobile:
image
https://github.com/user-attachments/assets/e9171048-4f89-4a69-b1a9-c9c4bf1f8ed5


components/HomepageSection/PowersAIAgentEconomies.jsx
```diff
@@ -0,0 +1,578 @@
...
+                    />
+                  )
+                );
+              })()}
+              <span>{activeTab}</span>
+            </div>
+            <ChevronDown size={24} />
+          </button>
+
+          {openDropdown && (
```
Tanya-atatakai, line 468: do you think it is possible to split mobileDropdown and desktop view into two components? and render like we do with `Menu` and `MenuMobile` - it will be easier to read

Tanya-atatakai (commented):
looks amazing! great work! Have left some comments, also noticed for me the arrows on the right have some gaps, probably due to fixed height. is it possible to fix?
image

PR #299, opened 2025-06-25
feat: add total olas staked

Tanya-atatakai:
## Proposed changes

1) make each card a component, as they were copy pasted for desktop and mobile - in case an update is required, we'd need to ensure both are updated
2) add total olas staked

## Screenshots/Recordings

Screenshot 2025-06-25 at 21 06 46

## Things to keep in mind


components/HomepageSection/Activity.jsx
```diff
@@ -23,19 +28,27 @@ import { useMemo } from 'react';
...
+  const [
+    transactions,
+    agents,
+    olasStaked,
+    dailyActiveAgents,
+    a2aTransactions,
+    feeMetrics,
+  ] = await Promise.allSettled([
+    getTotalTransactionsCount(),
+    getTotalUniqueStakers(),
```
Tanya-atatakai, line 40: replaced getTotalUnitsCount with this one, it's more correct

PR #311, opened 2025-07-10
chore: replaced all subsite links

atepem:
## Proposed changes

Replaced all links that have `olas.network` from using `<ExternalLink>` tag to new `<SubsiteLink>` tag.


atepem:
Have also made the whitepapers create a new tab when clicked as they come out as pdfs (have discussed this with Roman)

Tanya-atatakai (approved):

PR #313, opened 2025-07-14
style: misc updates and agentsfun

atepem:
## Proposed changes

* Fixed /operate link
* CTA and hero changes for agents.fun on both agents/ and agent-economies/
* added redirect from #get-involved section to /olas-token#choose-your-role

## Screenshots/Recordings

image
image


components/AgentEconomies/AgentsFunPage/Hero.jsx
```diff
@@ -14,14 +14,14 @@ const HeroImage = () => (
   />
 );
 
-const RunAnAgent = () => (
-  <Button variant="default" size="xl" asChild className="grow max-md:w-full">
-    <a href="#get-started">Run an Agent</a>
-  </Button>
-);
+// const RunAnAgent = () => (
```
Tanya-atatakai, line 17: could remove or make a comment to uncomment once the agent is back to work

components/HeroSection.jsx
```diff
@@ -13,17 +14,29 @@ export const HeroSection = ({
   description,
   PrimaryButton,
   SecondaryButton,
+  isUnderConstruction,
```
Tanya-atatakai, line 17: it's now specified to business logic, we could call it something like "statusTag?: React.ReactNode"
and render as is if passed. with this approach we could possibly use this component in pages/agents/shorts.jsx

pages/agents/agentsfun.jsx
```diff
@@ -133,50 +136,44 @@ const funAgents = [
     title: 'Agent Base',
     anchor: 'agent-base',
     description:
-      'Run your AI influencer on Base. Agent Base offers all the features you love — tweeting, evolving, benefits from meme coins, and interacting autonomously — all powered by the Olas. Get started today & create AI influencer on Base.',
+      'Run your AI influencer on Base. Agent Base offers all the features you love — tweeting, evolving, benefits from meme coins, and interacting autonomously — all powered by the Olas.',
     imgSrc: 'agentsfun-base.png',
-    link: '/pearl#download',
   },
-  // {
```
Tanya-atatakai, line 140: let's keep it commented out, even though we have the config in Pearl, we never pushed it live, so no need to show it as under construction

DavidMinarsch (approved):

PR #334, opened 2025-08-13
feat: replace dune metrics with subgraphs

Tanya-atatakai:
### metrics view:
Screenshot 2025-08-13 at 19 09 58

### Data explanation
Screenshot 2025-08-13 at 19 09 47
Screenshot 2025-08-13 at 19 09 41
Screenshot 2025-08-13 at 19 09 33

## Things to keep in mind


pages/api/main-metrics.js
```diff
@@ -0,0 +1,207 @@
...
+    );
+    const optimismPerformances = get7DaysPerformanceArray(
+      optimismResult.dailyActiveMultisigs_collection,
+    );
+
+    const totalAverage =
+      get7DaysAverage(gnosisPerformances) +
+      get7DaysAverage(basePerformances) +
+      get7DaysAverage(modePerformances) +
+      get7DaysAverage(optimismPerformances);
```
Tanya-atatakai, line 70: @atepem I realised it's done not quite correct on mech/bbdg pages
1) we should get average per 7 days pear each chain, and then get total for all chains, currently we get average across all chains which reduces the value (imagine we have 10 daas on gnosis, and 20 daas on base, our value will be 10 + 20 / 2 = 15, while in fact overall average value is 30)
2) in case on some days we didn't have any activity, we should count that towards the average correctly as 0, currently if from the query we only get activity for 3 days, we will do sum(activeMultisigCount) / 3, while we should divide by 7

please can you create a ticket for next week fixing this for bbdg and mech

atepem: Ok thanks, I'll add it on

components/DataPage/PredictAccuracy.jsx
```diff
@@ -23,10 +23,10 @@ export const PredictAccuracyInfo = () => {
           performance remains relevant.
         </p>
 
-        <p>The following queries are used:</p>
+        <p>The following queries is used:</p>
```
atepem, line 26: queries is -> query is

atepem (dismissed):
Looks good! Thanks for the changes and have added the task requested for next cycle :) One small comment and also,, the subgraph queries seem to take a while? Is this expected behaviour?

components/HomepageSection/Activity.jsx
```diff
@@ -26,29 +21,28 @@ import { useMemo } from 'react';
 const imgPath = '/images/homepage/activity/';
 
 const fetchMetrics = async () => {
```
rajat2502, line 23: should we move this into a hook?

pages/api/main-metrics.js
```diff
@@ -0,0 +1,207 @@
...
+    };
+
+    return data;
+  } catch (error) {
+    console.error('Error fetching main metrics:', error);
+    return null;
+  }
+};
+
+export default async function handler(req, res) {
```
rajat2502, line 181: really neat logic ✨

rajat2502 (dismissed):

components/DataPage/PredictAccuracy.jsx
Tanya-atatakai: ahhh :(

atepem (approved):

PR #349, opened 2025-09-03
chore: change how bbdg and mech 7 day avgs work

atepem:
## Proposed changes

Change how bbdg and mech avg agent activities are calculated:
* Avgs are calculated from each chain
* Added together as a total

## Screenshots/Recordings

Before:
image
image

After:
image
image

### DIFF
--- a/common-util/calculate7DayAverage.js
+++ b/common-util/calculate7DayAverage.js
@@ -1,8 +1,7 @@
-export const calculate7DayAverage = (performances) => {
-  if (performances.length === 0) return 0;
-  const total = performances.reduce(
-    (sum, p) => sum + Number(p.activeMultisigCount ?? 0),
-    0,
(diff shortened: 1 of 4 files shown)


common-util/calculate7DayAverage.js
```diff
@@ -0,0 +1,8 @@
+export const calculate7DayAverage = (performances) => {
```
Tanya-atatakai, line 1: can you please reuse it in `pages/api/main-metrics.js` instead of get7DaysAverage? I also think `get7DaysPerformanceArray` can be removed from there since we divide by 7 (so missing days will be count as zero anyway), and then we can pass a second param called "key" to understand by which key get the average (activeMultisigCount or count)

DavidMinarsch (approved):

PR #359, opened 2025-09-16
fix: hero section horizontal scroll bug

abhinav-anshul:
## Proposed changes

This PR fixes the existing horizontal scroll at the hero section in the live website

## Screenshots/Recordings
Screenshot 2025-09-16 at 19 43 45
If you see closely above, the scrollbar is present with animation being "cut" on the left side

&nbsp;
&nbsp;
&nbsp;
&nbsp;

Screenshot 2025-09-16 at 19 45 06
After the fix, there is no horizontal bar on the hero section, hence background animation being fixed as well

## Issue
The issue happens due to `w-full h-full` being paired with `100vw` which results in irregular bg animation as shown in the screenshot.

## Fix
I have added `inset-0` that will allow to cover the hero/background section being end to end

## Test
- tested on chromium
- tested on safari

### DIFF
--- a/components/HomepageSection/Hero.jsx
+++ b/components/HomepageSection/Hero.jsx
@@ -87,7 +87,7 @@ const Hero = () => (
     >
       <source src="/videos/homepage/hero-bg.webm" type="video/webm" />
     </video>
-    <div className="absolute w-full h-full overflow-hidden">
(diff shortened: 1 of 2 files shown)


rajat2502:
Thanks for raising. We have handled a ticket internally to handle this. Should be fixed in an upcoming PR.

PR #360, opened 2025-09-19
feat: add new GraphQL queries for per-agent onchain requests

axatbhardwaj:
## Proposed changes

Enabled classified request count from subgraphs for mech agent economies page

## Screenshots/Recordings

image

## Things to keep in mind

…etplace requests metrics…etplace requests metrics

### DIFF
--- a/pages/api/mech-metrics.js
+++ b/pages/api/mech-metrics.js
@@ -121,14 +121,8 @@ const fetchCategorizedRequestTotals = async () => {
     const addCounts = (records) => {
       if (!Array.isArray(records)) return;
       records.forEach((item) => {
-        const rawAgentId = item?.id;
-        let agentId;
-        if (typeof rawAgentId === 'number') {
-          agentId = rawAgentId;
-        } else if (typeof rawAgentId === 'string') {
-          const match = rawAgentId.match(/\d+/);
-          agentId = match ? Number(match[0]) : NaN;
-        }
+        if (!item?.id) return;
+        const agentId = Number(item.id);
         if (!Number.isFinite(agentId)) return;
         const requestCount = Number(item?.requestsCount ?? 0);
         combinedCounts.set(


components/DataPage/MechCategorizedRequests.jsx
```diff
@@ -0,0 +1,59 @@
...
+        <ExternalLink href={process.env.NEXT_PUBLIC_BASE_MM_SUBGRAPH_URL}>
+          Base
+        </ExternalLink>
+      </p>
+      <CodeSnippet>{MARKETPLACE_QUERY}</CodeSnippet>
+    </div>
+  </SectionWrapper>
+);
+
+
```
rajat2502, line 59:
```suggestion

```

pages/api/mech-metrics.js
```diff
@@ -91,6 +92,92 @@ const fetchMechRequestsFromSubgraphs = async () => {
   }
 };
 
+// Classified totals derived from subgraphs (Mech + Mech-Marketplace Gnosis + Base)
+const fetchCategorizedRequestTotals = async () => {
+  // Agent IDs by category (from SQL classification)
+  const predictTraderIds = [14, 25, 13];
+  const contributeIds = [6];
+  const governatooorIds = [5];
```
Tanya-atatakai, line 100: please let's move to constants similar to PREDICT_AGENT_CLASSIFICATION, and reuse in components/DataPage/MechCategorizedRequests.jsx

pages/api/mech-metrics.js
```diff
@@ -91,6 +92,92 @@ const fetchMechRequestsFromSubgraphs = async () => {
   }
 };
 
+// Classified totals derived from subgraphs (Mech + Mech-Marketplace Gnosis + Base)
+const fetchCategorizedRequestTotals = async () => {
+  // Agent IDs by category (from SQL classification)
```
Tanya-atatakai, line 97: sorry what is sql calssification?

pages/api/mech-metrics.js
```diff
@@ -91,6 +92,92 @@ const fetchMechRequestsFromSubgraphs = async () => {
...
+          mechMarketplaceRequestsPerAgentsQuery(allIds.map(String)),
+        ),
+        ATA_GRAPH_CLIENTS.base.request(
+          mechMarketplaceRequestsPerAgentsQuery(allIds.map(String)),
+        ),
+      ]);
+
+    const combinedCounts = new Map();
+
+    const addCounts = (records, idFieldName, countFieldName) => {
```
Tanya-atatakai, line 119: the `idFieldName` and `countFieldName` are the same correspondingly - `id` and `requestsCount`. maybe we can remove the parameters?

pages/api/mech-metrics.js
```diff
@@ -91,6 +92,92 @@ const fetchMechRequestsFromSubgraphs = async () => {
...
+      if (!Array.isArray(records)) return;
+      records.forEach((item) => {
+        const rawAgentId =
+          item?.[idFieldName] !== undefined
+            ? item?.[idFieldName]
+            : (item?.agentId ?? item?.id);
```
Tanya-atatakai, line 125: so with the passed idFieldName as `id` it'll be:
```
          item?.id !== undefined
            ? item?.id
            : (item?.agentId ?? item?.id);
```

pages/api/mech-metrics.js
axatbhardwaj: Done

components/DataPage/MechCategorizedRequests.jsx
```diff
@@ -0,0 +1,72 @@
...
+    requestsCount
+  }
+}`;
+
+const MARKETPLACE_QUERY = `query {
+  requestsPerAgents(where: { id_in: ${IDS_LIST} }) {
+    id
+    requestsCount
+  }
+}`;
```
Tanya-atatakai, line 26: we can simply import from graphql/queries, no? also would be no need to map and stringify

pages/api/mech-metrics.js
```diff
@@ -91,6 +95,76 @@ const fetchMechRequestsFromSubgraphs = async () => {
...
+      records.forEach((item) => {
+        const rawAgentId = item?.id;
+        let agentId;
+        if (typeof rawAgentId === 'number') {
+          agentId = rawAgentId;
+        } else if (typeof rawAgentId === 'string') {
+          const match = rawAgentId.match(/\d+/);
+          agentId = match ? Number(match[0]) : NaN;
+        }
+        if (!Number.isFinite(agentId)) return;
```
Tanya-atatakai, line 132: can you please give me an example of a string agent id that represents not a number?
in both marketplace and legacy mech I just see such ids like `"12"`, or `"5"` etc. so why should we complicate the logic here instead of just simply doing the following
```suggestion
      records.forEach((item) => {
        if (item?.id) return;
        const agentId = Number(item.id);
```

PR #372, opened 2025-10-07
Feature/olas token

axatbhardwaj:
## Proposed changes

Using Tokenomics subgraphs for Fetching total Holders from Olas-token

## Screenshots/Recordings

image

## Things to keep in mind


atepem (approved):
Is it maybe my cache but I'm not able to see the tokenomics results on the page?

axatbhardwaj:
> Is it maybe my cache but I'm not able to see the tokenomics results on the page?

because currently the ENV vars are not set @atepem , as these subgraphs are never used before this in the website

.env.example
axatbhardwaj, line 1: @Tanya-atatakai , we need these new ENV vars :) 🙇

atepem:
Also (just suggestion), do you think it might be a good idea to make the data page content into a json + repeating component since each section appears to have the same layout (id, title, code snippet, links and descriptions?).
Also maybe a contents part as the page seems to be getting longer and longer (or a return to top button).

axatbhardwaj:
> Also (just suggestion), do you think it might be a good idea to make the data page content into a json + repeating component since each section appears to have the same layout (id, title, code snippet, links and descriptions?). Also maybe a table of contents part as the page seems to be getting longer and longer (or a return to top button). But this is not necessary to add in this iteration.

sure would keep it in mind from next change

common-util/api/tokenomics.js
```diff
@@ -0,0 +1,36 @@
...
+const toLowerCaseAddress = (address) => address.toLowerCase();
+
+const fetchHolderCount = ({ key, token }) => {
+  const client = TOKENOMICS_GRAPH_CLIENTS[key];
+  if (!client) {
+    return Promise.resolve(0);
+  }
+
+  return client
+    .request(holderCountsQuery, { tokenId: toLowerCaseAddress(token) })
```
Tanya-atatakai, line 14: we shouldn't lowercase address if the type is Bytes in subgraph, checksummed address should work as well. could you please check?

common-util/api/tokenomics.js
```diff
@@ -0,0 +1,36 @@
...
+      return total;
+    }
+
+    return Number.isFinite(result.value) ? total + result.value : total;
+  }, 0);
+
+export const getTotalTokenHolders = async () => {
+  const results = await collectHolderCounts();
+  return sumHolderCounts(results);
+};
```
Tanya-atatakai, line 36: I think it's a bit overkill to make these small functions for such simple logic

common-util/constants.js
```diff
@@ -81,6 +79,28 @@ export const ACCELERATOR_APPLY_URL =
...
+export const TOKEN_HOLDER_NETWORKS = [
+  { key: 'ethereum', token: '0x0001A500A6B18995B03f44bb040A5fFc28E45CB0' },
+  { key: 'arbitrum', token: '0x064F8B858C2A603e1b106a2039F5446D32dc81c1' },
+  { key: 'base', token: '0x54330d28ca3357F294334BDC454a032e7f353416' },
+  { key: 'celo', token: '0xaCFfAe8e57Ec6E394Eb1b41939A8CF7892DbDc51' },
+  { key: 'gnosis', token: '0xcE11e14225575945b8E6Dc0D4F2dD4C570f79d9f' },
+  { key: 'optimism', token: '0xFC2E6e6BCbd49ccf3A5f029c79984372DcBFE527' },
+  { key: 'polygon', token: '0xFEF5d947472e72Efbb2E388c730B7428406F2F95' },
+  { key: 'mode', token: '0xcfD1D50ce23C46D3Cf6407487B2F8934e96DC8f9' },
+];
```
Tanya-atatakai, line 91: can we somehow reuse data from data/tokens.json?

Tanya-atatakai (changes requested):

common-util/api/tokenomics.js
Tanya-atatakai, line 1: this should be moved to the pages/api instead, otherwise we expose the api key

components/OlasTokenPage/tokenMetrics.js
```diff
@@ -0,0 +1,37 @@
+import { TOKEN_HOLDER_NETWORKS } from 'common-util/constants';
```
rajat2502, line 1: I didn't understand why two different files tokenMetrics/tokenomics?

PR #385, opened 2025-10-29
Feature/pearl daa

axatbhardwaj:
## Proposed changes

- Using the Predict Daa's for Pearl only using staking contracts and checkpoint inference 
- we use predict-pearl-DAA and optimus DAA's (all qs+pearl) since there is no way to distinguish ATM
- updated datapage for the same

## Screenshots/Recordings

image

image

## Things to keep in mind

### DIFF
--- a/components/DataPage/DailyActiveAgents.jsx
+++ b/components/DataPage/DailyActiveAgents.jsx
@@ -154,9 +154,10 @@ export const DailyActiveAgentsInfo = () => {
         <div className="space-y-6 mt-4">
           <p>
             Tracks how many unique multisigs were active each day for all agents
-            in the Pearl app. The <strong>active multisig count</strong> reflects
-            number of unique multisigs that performed at least one on-chain
-            interaction attributed to these agents within the UTC day window.
+            in the Pearl app. The <strong>active multisig count</strong>{' '}
+            reflects the number of unique multisigs that performed at least one
+            on-chain interaction attributed to these agents within the UTC day
+            window.
           </p>
 
           <p>


pages/api/pearl-daa.js
```diff
@@ -0,0 +1,216 @@
...
+    });
+    return acc;
+  }, {});
+};
+
+const getPerformanceByAgentId = async () => {
+  // Only use valory_trader agent IDs
+  const traderAgentIds = PREDICT_AGENT_CLASSIFICATION.valory_trader.map((n) =>
+    Number(n),
+  );
```
Tanya-atatakai:
```suggestion
  const traderAgentIds = PREDICT_AGENT_CLASSIFICATION.valory_trader
```

those are already numbers

pages/api/pearl-daa.js
```diff
@@ -0,0 +1,216 @@
...
+
+    performances.forEach((perf) => {
+      // Extract timestamp from id: "day-{timestamp}-agent-40"
+      const timestampMatch = perf.id.match(/day-(\d+)-agent-40/);
+      if (!timestampMatch) {
+        console.error('Invalid id format:', perf.id);
+        return;
+      }
+      const timestamp = Number(timestampMatch[1]);
+      const date = new Date(timestamp * 1000).toISOString().slice(0, 10);
```
Tanya-atatakai: please can you tell why do we need milliseconds

pages/api/pearl-daa.js
```diff
@@ -0,0 +1,216 @@
...
+};
+
+const calculate7DayTrailingAverage = (dailyCounts) => {
+  if (dailyCounts.length < 7) {
+    return 0;
+  }
+
+  const sum = dailyCounts.reduce((acc, count) => acc + count, 0);
+  return sum / 7;
+};
```
Tanya-atatakai: please reuse calculate7DayAverage from `common-util/calculate7DayAverage.js`

pages/api/pearl-daa.js
```diff
@@ -0,0 +1,216 @@
...
+
+const getCombinedPearlDAAs = async () => {
+  const [pearlDailyCounts, optimusDailyCounts] = await Promise.all([
+    getPearlDailyCounts(),
+    getOptimusDailyCounts(),
+  ]);
```
Tanya-atatakai: should be called
```suggestion
  const [pearlDailyCounts, optimusDailyCounts] = await Promise.all([
    getPredictDailyCounts(),
    getBabydegenOptimusDailyCounts(),
  ]);
```

pages/api/pearl-daa.js
```diff
@@ -0,0 +1,216 @@
...
+
+  const allDates = new Set([
+    ...Object.keys(pearlDailyCounts),
+    ...Object.keys(optimusDailyCounts),
+  ]);
```
Tanya-atatakai: please use `Promise.allSettled` for the above, if one of the calls fails, we won't see any data

pages/api/pearl-daa.js
```diff
@@ -0,0 +1,216 @@
...
+
+  const sortedDates = Array.from(allDates).sort().slice(-7);
```
Tanya-atatakai: if the calculation is correct, the array should be of length = 7, no need to slice

pages/api/pearl-daa.js
```diff
@@ -0,0 +1,214 @@
...
+
+    combinedCounts[dateString] = {
+      predict: predictCount,
+      babydegen: babydegenCount,
+      total: combined,
+    };
+    dailyCountsArray.push({ count: combined });
+  });
+
+  const averageDAAs = calculate7DayAverage(dailyCountsArray, 'count');
```
Tanya-atatakai: I think we're doing wrong math here.
we have to get 7 days average DAA for predict and separately for babydegen. then sum up the averages.
we also shouldn't sort timestamps and make some new dateString etc., this is confusing and not needed.
as well as we shouldn't do `groupServicesByDays`, we already have them by days at least for babydegen count. for predict we only need to update total counter and keep there only pearl contracts from checkpoints.

please let's make it simple
```suggestion
  const babydegenAverage = calculate7DayAverage(dailyAgentPerformances, 'activeMultisigCount'); // dailyAgentPerformances are just performances from getBabydegenOptimusDailyCounts
  const predictAverage = calculate7DayAverage(dailyAgentPerformances, 'activeMultisigCount'); // dailyAgentPerformances is updated based on the logic where we should exclude non-pearl
  
  const averageDAAs = babydegenAverage + predictAverage;
```

you can see an example here as well: https://github.com/valory-xyz/olas-website/blob/main/pages/api/babydegen-metrics.js#L335-L344

Tanya-atatakai (changes requested):

Tanya-atatakai (approved):

common-util/api/index.js
```diff
@@ -133,6 +133,21 @@ export const getPredictMetrics = async () => {
   }
 };
 
+// ----------- PEARL DAAS -----------
+export const getPredictDAAs = async () => {
```
rajat2502:
```suggestion
export const getPearlDAAs = async () => {
```

rajat2502 (approved):

DavidMinarsch (approved):

PR #389, opened 2025-11-04
fix: copy to clipboard option

rajat2502:
## Proposed changes

Fix random fn and add copy to clipboard option.

## Things to keep in mind


pages/card/index.jsx
```diff
@@ -199,22 +242,99 @@ const CommunityCardClient = () => {
...
+      setToast({
+        message: 'Image copied to clipboard!',
+        type: 'success',
+      });
+    } catch (e) {
+      console.error('Failed to copy image to clipboard:', e);
+      setToast({
+        message: 'Copying to the clipboard failed, please download the card.',
+        type: 'error',
+      });
```
mohandast52: Repeated thrice, could be just a reusable function.

atepem (approved):

PR #424, opened 2026-01-04
Add link to full Mech metrics page from marketplace

DavidMinarsch:


components/MechMarketplacePage/MarketplaceMetrics.jsx
```diff
@@ -56,6 +57,27 @@ export const MarketplaceMetrics = () => {
       {marketplaceData.map((data, index) => (
         <MetricsCard key={index} metrics={data} />
       ))}
+      <div className="mt-8 text-center">
+        <Link
+          href="/agent-economies/mech"
+          className="inline-flex items-center text-purple-600 hover:text-purple-700 font-medium transition-colors"
+        >
+          More metrics on Mech Marketplace
+          <svg
```
atepem: Small: may be good to use `ChevronRight` from lucide-react here just to stay consistent and avoid unnecessary custom svgs

PR #436, opened 2026-01-08
[BDMARK 1542] feat: website metrics guardrails

rajat2502:
## Proposed changes

Add logic to mark metrics as stale in case any of the subgraphs (or dune) errors out. In the event of errors, we will display the "last known good" value of the metric and gray out the metric number. We would also show a tooltip mentioning the timestamp for the good value and the sources that caused the issue.

Screenshot 2026-01-09 at 2 58 57 PM

This was the first time I created the blob, and this metric failed, which is why it shows "Unknown" for the timestamp. Ideally, we would have a proper time value. 

## Things to keep in mind

### DIFF
--- a/common-util/api/predict/index.ts
+++ b/common-util/api/predict/index.ts
@@ -64,6 +64,7 @@ const fetchPredictDaa7dAvg = async (): Promise<
       timestamp_gt,
       timestamp_lt,
     },
+    // TODO: move these to consts for better maintainability
     source: 'registry:gnosis',
     transform: (data) => {
       const rows = data.dailyAgentPerformances || [];


README.md
```diff
@@ -62,6 +62,28 @@ The `functions` block configures resource limits for serverless API routes:
...
+
+#### Breaking Changes
+
+> **⚠️ Important:** When making breaking changes to the metrics schema (adding/removing/renaming fields), you **must update the blob prefix** to create a new version of the blobs.
+
+During ISR, Next.js expects the stored blob schema to match what the page components expect. If you change the schema without updating the prefix, ISR pages may fail to render due to schema mismatches.
+
+**To introduce a new blob version:**
+
+1. Update the prefix in `snapshot-storage.ts` (e.g., `metrics-v2-${process.env.NODE_ENV}`)
```
mohandast52: what about today's date?
```suggestion
1. Update the prefix in `snapshot-storage.ts` (e.g., `metrics-2026-01-30-${process.env.NODE_ENV}`)
```

common-util/snapshot-storage.ts
```diff
@@ -11,6 +14,97 @@ type SaveSnapshotParams = {
...
+    return newData;
+  }
+
+  if ('value' in newData && 'status' in newData) {
+    const newMetric = newData as MetricWithStatus<unknown>;
+    const oldMetric = oldData as MetricWithStatus<unknown>;
+
+    if (
+      newMetric.value === null ||
+      newMetric.value === undefined ||
```
mohandast52: `isNil` from lodash?

common-util/api/main-metrics.ts
```diff
@@ -15,359 +16,429 @@ import {
...
+  const chains = [
+    'gnosis',
+    'base',
+    'mode',
+    'optimism',
+    'celo',
+    'ethereum',
+    'polygon',
+    'arbitrum',
+  ] as const;
```
mohandast52: can be reused?

common-util/api/main-metrics.ts
```diff
@@ -15,359 +16,429 @@ import {
...
-            (feeValue as MechFeesResult['global']).totalFeesInUSD || '0'
-          );
-          totalFees += usdValue;
+      const source = sources[index];
+      if (result.status === 'rejected') {
+        console.error(`mechFees:${source}`, result.reason);
+        fetchErrors.push(`mechFees:${source}`);
+      } else {
+        const data =
+          index === 2
```
mohandast52: what does index === 2 mean?

common-util/api/other-metrics/build.ts
```diff
@@ -1,25 +1,32 @@
...
-    console.error('Error fetching total builders:', error);
-    return null;
-  }
+const fetchTotalBuilders = async (): Promise<
+  MetricWithStatus<number | null>
+> => {
+  return executeGraphQLQuery<TotalBuildersResult, number | null>({
+    client: autonolasGraphClient,
+    query: totalBuildersQuery,
+    source: 'build:totalBuilders',
```
mohandast52: Are we planning to make it const or keep it as a string everywhere?

mohandast52 (approved):

common-util/api/other-metrics/build.ts
rajat2502: we can make it a const, i'll add a todo.

PR #437, opened 2026-01-08
style: (BDMARK-1535)  Update Olas Roadmap page based on YiR blog

atepem:
## Proposed changes

Update roadmap on desktop and mobile.
* Text
* background
* table

## Screenshots/Recordings

Screenshot 2026-01-08 164940
Screenshot 2026-01-08 164904

### DIFF
--- a/components/RoadmapPage/MobileTable.tsx
+++ b/components/RoadmapPage/MobileTable.tsx
@@ -15,7 +15,7 @@ const TableCell = ({
   const colorClass =
     column === 'pearl' ? 'roadmap-cell-pearl' : 'roadmap-cell-marketplace';
   return (
-    <td className={`border border-slate-200 p-3 ${colorClass}`}>
+    <td className={`border border-slate-300 p-3 ${colorClass}`}>
       {showCheckmark && <CheckCheck className="inline mr-2" />}
       {children}
     </td>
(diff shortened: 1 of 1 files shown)


components/RoadmapPage/MobileTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+        <tr>
+          <th className="w-1/2 font-medium border border-slate-200 p-3">
+            Pearl
+          </th>
+          <th className="w-1/2 font-medium border border-slate-200 p-3">
+            Marketplace
+          </th>
+        </tr>
+      </thead>
+      <tbody>
```
Tanya-atatakai: would be a bit simpler to read and modify in future if we made it a configurable list of items on which we then do .map and paint first item in a row as `column="pearl"` and second `column="marketplace"`

components/RoadmapPage/MobileTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+export const MobileTable = () => (
+  <div className="md:hidden w-full h-[1420px] border border-t-1.5 mobile-roadmap-container relative">
+    <table className="table-fixed w-full border border-slate-200 mobile-roadmap-table">
+      <thead className="bg-slate-100 text-center text-lg text-black">
+        <tr>
+          <th className="w-1/2 font-medium border border-slate-200 p-3">
+            Pearl
+          </th>
+          <th className="w-1/2 font-medium border border-slate-200 p-3">
```
Tanya-atatakai: let's replace `border-slate-200` with `border-slate-300`, this is closer to #C2CBDB that is in figma (as per chatgpt)

atepem: ohhh, i used this tool to find closer colours https://find-nearest-tailwind-colour.netlify.app/

Tanya-atatakai (dismissed):

PR #453, opened 2026-02-02
feat: add polygon support to metrics

Tanya-atatakai:
## Proposed changes

1) add missing envs to env.example, remove unused ones + the related code & package (we download Pearl assets from Pearl.you now and it's not needed)
2) add polygon to where mech and staking subgraphs are used
3) update data page so that all links are used from envs and use `getSubgraphExplorerUrl` that converts the query link to explorer link

## Things to keep in mind


rajat2502 (approved):
LGTM!
Thanks for the refactor. 
Also, because we have updated a few query functions, the results wouldn't directly reflect when we go live, they would take upto 6 hrs (for agent-economies and other metrics, while 1 hour for main-metrics). I hope that's okay 👀? In order to test, you can run this locally and it would only update the development blob

PR #488, opened 2026-04-14
fix(predict): correct mech attribution and prevent QMR leak in ROI Distribution

LOCKhart07:
## Summary

Two data-affecting fixes to the **Polystrat / Omenstrat Partial ROI Distribution** chart pipeline (`common-util/api/predict/roi-distribution.ts`). Spec aligned with @Tatiana's feedback.

### What's fixed

1. **QMR TTL age-out (14 days).** The actual bug. Pending mech-request entries whose titles never settled (invalid markets, schema drift, title truncation) were leaking forever and being subtracted from each agent's lifetime mech-fee count. Net effect: Max-tab ROI systematically **inflated** for ~87% of agents. Fix: orphaned entries expire at 14 days and count as settled.

2. **Activity threshold (`MIN_TRADES_FOR_ROI_DISPLAY = 10`).** Mirrors the trader skill's "need more data" rule. Agents with < 10 lifetime bets are excluded from every tab. Replaces the prior `roi === -100` silent drop, which was both a workaround for a real subgraph issue *and* an implicit filter against low-activity noise — the new rule is principled and handles both concerns: real -100% agents with enough activity still appear; 1-bet wallets don't pollute the tail.

Internal change: `QmrData.questionMechRequests` schema is now `Record<string, Record<string, number[]>>` (timestamp arrays) so TTL can know request ages. Attribution of mech requests to the settlement day is unchanged — per product intent, all market costs land on the day the market settles, not the day the request was made.

### Audit findings (Polystrat, real subgraph data)

| | Before | After | Δ |
|---|---:|---:|---:|
| QMR pending titles | 2,450 | 777 | **−1,673** |
| Stale entries flushed (one-time) | — | **23,147** | — |
| Mean per-agent ROI shift (Max tab, qualifying agents) | — | −0.34 pp | 122/123 downward |
| Max abs ROI shift | — | −1.32 pp | — |

**Agents included per tab (after ≥10-bet filter):**

| Tab | Included | Excluded (low activity) |
|---|---:|---:|
| 7D  | 33  | 80 |
| 30D | 115 | 31 |
| 90D | 129 | 29 |
| Max | 129 | 20 |

**Agents at exactly −100% ROI making it into the histogram:** 0 (all filter out for low activity). The alarming left bar from v1 of this PR is gone.

### Storage impact

QMR blob shrinks ~3× (2,450 → 777 titles) on first run after merge. Schema for `QmrData.questionMechRequests` changes from `Record<string, Record<string, number>>` (counts) to `Record<string, Record<string, number[]>>` (timestamp arrays). Existing snapshots will be re-built from genesis on first refresh — handled by the existing incremental-update logic since `lastMechRequestTimestamp` is preserved separately.

### Queries changed

`getOmenTraderAgentsQuery` and `getPolymarketTraderAgentsQuery` now also request `totalBets` (already present on both subgraph schemas — verified).

### Out of scope (follow-up PRs)

- FIX-3: surface `_meta.hasIndexingErrors` per query + stale badge in UI
- FIX-4: confirm `dailyProfit` subgraph definition + sanity guard
- FIX-5: chart copy / tab labels / footnotes
- FIX-7: bump `BYDAY_RETENTION_DAYS` from 90 → 95
- FIX-8 / 9 / 10 / 11: config externalisation, collision logging, bin-label clarity, cross-repo metric docs

## Test plan

- [ ] Trigger `/api/refresh-metrics/predict-roi-distribution?agent=polystrat` in preview environment. Expect first run to take ~20s with a `[roi-dist:polystrat] expired N QMR entries older than 14 days` log line.
- [ ] Repeat for `?agent=omenstrat`.
- [ ] Verify the Polystrat Partial ROI Distribution chart renders on the Predict page across all four tabs (7D / 30D / 90D / Max). No `−100% to −90%` bar (filter excludes those agents).
- [ ] Spot-check 2-3 individual agent ROIs vs the trader-side `/api/v1/agent/performance` Lifetime partial ROI to confirm directional alignment.
- [ ] Check Vercel function logs for `[roi-dist:polystrat:<tab>] included=X, excluded_low_activity=Y` on every histogram compute.

### DIFF
--- a/common-util/api/predict/roi-distribution.ts
+++ b/common-util/api/predict/roi-distribution.ts
@@ -20,6 +20,8 @@ const MAX_DAYS_PER_RUN = 30;
 const DAY_SECONDS = 86400;
 // Keep only this many days in byDay (covers the longest non-global tab: 90D)
 const BYDAY_RETENTION_DAYS = 90;
+// FIX-2: age-out TTL for pending QMR entries (markets resolve in ~4 days)
+const QMR_MAX_AGE_DAYS = 14;
 
 // Genesis timestamps (UTC midnight) for each agent type
 const OMEN_GENESIS_TS = 1763769600;
@@ -27,6 +29,8 @@ const POLYMARKET_GENESIS_TS = 1768867200;
 // Earliest block timestamp to consider when fetching mech requests
 const GNOSIS_MECH_REQUESTS_GENESIS_TS = 1763078400;
 const POLYGON_MECH_REQUESTS_GENESIS_TS = 1763078400;
+
+const dayKeyOf = (ts: number): string => String(Math.floor(ts / DAY_SECONDS) * DAY_SECONDS);
 
 // ─── Blob related query types ────────────────────────────────────────────────────
 
@@ -39,8 +43,13 @@ const POLYGON_MECH_REQUESTS_GENESIS_TS = 1763078400;
  * For such markets its entry is deleted from QMR and the count is materialized
  * into byDay[date].agents[agentId].participants.
  */
+/**
+ * FIX-1: Store per-request blockTimestamps (ascending) instead of a flat count.
+ * On settlement, each request is attributed to the day it was actually made,
+ * instead of being collapsed onto the settlement day.
+ */
 export type QmrData = {
-  questionMechRequests: Record<string, Record<string, number>>; // title → agentId → count
+  questionMechRequests: Record<string, Record<string, number[]>>; // title → agentId → sorted asc timestamps
   lastMechRequestTimestamp: number;
 };
 
@@ -177,8 +186,9 @@ const fetchPolystratDailyStats = async (
 const fetchIncrementalMechRequests = async (
   chain: 'gnosis' | 'polygon',
   lastTimestamp: number
-): Promise<{ additions: Record<string, Record<string, number>>; lastTimestamp: number }> => {
-  const additions: Record<string, Record<string, number>> = {};
+): Promise<{ additions: Record<string, Record<string, number[]>>; lastTimestamp: number }> => {
+  // FIX-1: additions now stores timestamps per (title, agentId), not just counts.
+  const additions: Record<string, Record<string, number[]>> = {};
   let latestTs = lastTimestamp;
   let skip = 0;
   const client = MARKETPLACE_GRAPH_CLIENTS[chain];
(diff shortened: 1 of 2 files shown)


LOCKhart07:
image

mariapiamo (changes requested):
Findings (Ordered by Severity)

P0: Backward-incompatible QMR schema can crash refresh at runtime.
Risk points: roi-distribution.ts (line 385), roi-distribution.ts (line 387), roi-distribution.ts (line 504).
Cause: old blobs may still store numeric QMR values, while new code expects arrays.
P1: Legacy numeric QMR values are miscounted as zero pending.
Risk point: roi-distribution.ts (line 317).
Cause: tsList?.length returns undefined for numbers, so open requests are undercounted and settled mech requests are overstated.
Required Actions Before Merge

Add backward-compatible QMR value handling everywhere QMR is read:
Array.isArray(value) ? value.length : Number(value ?? 0).
Normalize legacy QMR entries before merge/TTL loops.
Add tests for mixed legacy/new QMR snapshot shapes.
If not fixing in code, perform explicit migration/reset of roi-distribution/*-requests blobs at deployment.

Strategic Suggestion (V2, simpler and safer) @LOCKhart07 @Tanya-atatakai 

Move ROI computation to an event-based daily ledger per agent:
mech_cost by request timestamp day.
market_cost and market_payout by resolution/settlement day.
Compute window ROI directly from ledger sums:
ROI = (payout - market_cost - mech_cost) / (market_cost + mech_cost).
Use simple “active in selected window” inclusion (optionally require at least 1 settled market).
Retire title-based QMR-to-market matching and orphan TTL compensation from ROI core path.
Why this V2 helps

Removes the most fragile logic (title association).
Eliminates orphan-fee ambiguity by construction.
Makes behavior easier to reason about, test, and maintain.

LOCKhart07:
Thanks @mariapiamo — addressed the P0/P1 in c818a0bf.

**Approach:** normalize-on-load rather than per-site shims. Added `normalizeQmrShape()` at the start of `updateAgentBlueprintData` that converts legacy `number` values into `number[]`. Downstream code (the spots flagged at 317, 385, 387, 504) only ever sees the new shape, so no dual-shape handling leaks into the TTL/merge/counting paths.

**Legacy timestamp choice:** legacy entries have no real per-request timestamps. Stamping with `lastMechRequestTimestamp` (preserved separately in the blob) bounds their real age from above — they can't be newer than that, so TTL still ages them out correctly and we avoid the alternative failure modes (stamping with `now` = 14-day artificial grace; stamping with `0` = everything flushes to epoch day 0 on first run).

**On the V2 ledger suggestion:** strongly agree it's the right long-term shape — title-based matching is the fragile layer that caused the original bug. I'd like to do that as a follow-up rather than expand this PR; the audit numbers in the description are scoped to the current pipeline and the fix is already blocking a visible chart regression. Happy to open a tracking issue for the ledger refactor if you want.

mariapiamo (approved):

PR #493, opened 2026-04-22
fix: compute bond-page POL and fees from liquidity subgraphs

kupermind:
## Summary

The `/bond` page was showing a wildly inflated Total Protocol-owned Liquidity (~**\$1.14 × 10²⁰** instead of ~\$2.5M) and fetched Fees from POL via Dune (stale: ~\$2.77M). This PR fixes both: POL is computed correctly, fees now come from the same liquidity subgraphs, and a sanity-clamp prevents this class of bug from ever persisting a broken value in the snapshot again.

Implementation mirrors [`autonolas-subgraph-studio/scripts/pol-aggregation.js`](https://github.com/valory-xyz/autonolas-subgraph-studio/blob/main/scripts/pol-aggregation.js) — the canonical off-chain reference — so the website and the script now agree.

## Root cause of the broken POL number

Two compounding bugs in `common-util/api/other-metrics/protocol.ts` + `liquidityL2Query`:

1. **Base has two pools** (OLAS-USDC and WETH-OLAS) in one subgraph. The query asked for `first: 1`, and the Base valuation formula blindly assumed OLAS-USDC: `reserve1 / 1e6 × 2`. When the subgraph returned the WETH-OLAS pool instead, `reserve1` was OLAS with **18 decimals** being read as USDC with **6 decimals** — inflating the value by ~10¹² per pool.
2. Even with both pools fetched, bridged-LP share lookup used `bridgedBalances['base']` for both, but the WETH-OLAS pool's `originChain` key is `'base-weth'`, so the WETH pool's Treasury share would have been `0`.

### 1. POL + fees computation (`common-util/api/other-metrics/protocol.ts`)

- Query `first: 10` pools per L2 subgraph.
- Dispatch Base by pool address — OLAS-USDC and WETH-OLAS each get their own valuation, fee-conversion, and origin-chain key.
- Single-pool chains (Gnosis / Polygon / Arbitrum / Optimism / Celo) keyed by chain name.
- Fees pulled from the same liquidity subgraphs:
  - Ethereum uses pre-scaled `cumulativeProtocolFeesUsd` directly.
  - L2s convert `cumulativeFeesToken0/1` → USD via the BigInt-safe `l2FeesToUsd` helper (avoids `Number` precision loss on 18-decimal cumulative values), then multiply by each pool's Treasury share.
- Solana unchanged (2 × SOL vault × SOL/USD × 99.995%, no fees tracked).

### 2. Sanity-clamp (defense-in-depth against decimals-mismatch bugs)

Two layers, added so a future decimals bug can't persist a garbage snapshot:

- **Per-pool**: any single pool with TVL > **\$500M** or cumulative fees > **\$100M** is skipped with a `fetchError`. Almost certainly indicates a decimals/pair mismatch rather than real growth.
- **Final total**: if aggregate POL > **\$1B** or cumulative fees > **\$500M**, the fetcher returns `null` instead of the value. `mergeWithFallback` then preserves the previous valid snapshot value rather than persisting the garbage.

Bounds are intentionally generous (~200× headroom over current POL of ~\$2.5M). Breaching them signals a bug, not organic growth.

### 3. Query updates (`common-util/graphql/queries.ts`)

- `liquidityEthQuery`: added `cumulativeFeesUsd`, `cumulativeProtocolFeesUsd`, `cumulativeExternalFeesUsd`.
- `liquidityL2Query`: `first: 10`, added `cumulativeFeesToken0/1`.

### 4. UI

- `BondsPage/BondMetrics.tsx`: Fees card now links to `/data#protocol-liquidity-fees` (internal) instead of the Dune query URL.
- `DataPage/ProtocolOwnedLiquidity.tsx`: rewrote the Fees section to describe the subgraph-based computation; removed the Dune Analytics reference.
- Dropped unused constants from `common-util/constants.ts`: `DUNE_QUERY_URL`, `DUNE_TOTAL_PROTOCOL_REVENUE_URL`, `TOTAL_PROTOCOL_REVENUE_FROM_FEES_ID`.

### 5. Docs

- `CLAUDE.md`: added a bullet under *Working with Blockchain Data* recording the sanity-clamp convention for cross-chain USD aggregation.

## Expected values

From `pol-aggregation.js --json` at the time of this PR, and confirmed end-to-end against the preview blob refresh output:

| Metric | Before | After |
| --- | --- | --- |
| Total POL | \$114,570,687,335,328,450,000 | **\$2,525,016** |
| Fees from POL | \$2,774,197 (Dune, stale) | **\$2,663,997** |

## DUNE_API_KEY is no longer needed

Dune was the only consumer of this key; this PR removes the last call to `api.dune.com`. Cleanup done here:

- Removed `DUNE_API_KEY=__KEY__` from `.env.example`.
- Removed the `Analytics: NEXT_PUBLIC_DUNE_API_KEY` line from `CLAUDE.md` (stale — the actual variable name was `DUNE_API_KEY`, and no code reads it).
- No code paths reference `DUNE_API_KEY` after this PR (verified via `grep -r DUNE_API_KEY`).

### Action required after merge

> **Please delete `DUNE_API_KEY` from the Vercel project environment variables** (Production, Preview, and Development scopes) — it is now unused.

The two remaining `dune.com` references in the repo (`data/kits.json`, `data/agents.json`) are external links to a community-published Dune dashboard (`dune.com/adrian0x/aks`) surfaced in the kits/agents lists — those are user-facing content, not API consumers, and are intentionally left alone.

## Snapshot behavior after merge

`/bond` reads from a Vercel Blob snapshot (`metrics-$NODE_ENV-other.json`) regenerated by the cron at `0 */6 * * *`. After merge, either:

- Wait for the next cron tick (up to 6 h), then wait up to 5 min for ISR on `/bond` to pick up the new snapshot, or
- Trigger a refresh manually: `curl https://<deployment>/api/refresh-metrics/other` (can take ~3 min — the endpoint fans out across 7 liquidity subgraphs + Solana RPC + other non-protocol fetchers, and is declared with `maxDuration: 300`).

## Test plan

- [ ] `yarn lint` clean.
- [ ] `yarn build` clean.
- [ ] Trigger `GET /api/refresh-metrics/other`; inspect JSON response confirms `totalProtocolOwnedLiquidity.value ≈ 2_525_016` and `totalProtocolRevenue.value ≈ 2_663_997`.
- [ ] Visit `/bond` after ISR window and confirm both cards render sane numbers.
- [ ] Confirm Fees card "View source" link goes to `/data#protocol-liquidity-fees`.
- [ ] Visit `/data#protocol-liquidity-fees` and confirm the description describes the subgraph math, with no Dune link.
- [ ] After merge: delete `DUNE_API_KEY` from the Vercel project environment.

## Before
image

## After
image


Tanya-atatakai:
Reviewed as senior FE. Direction is right: fewer external deps, atomic POL+fees fetching, sanity-clamp protects against decimals regressions. Comments below.

## Bugs / correctness

1. **Dead fields in the ETH query** — `common-util/graphql/queries.ts:942,944` and `protocol.ts:21,23` fetch and type `cumulativeFeesUsd` and `cumulativeExternalFeesUsd` but nothing reads them. Drop, or leave a comment saying why they're reserved.

2. **Missing `prices.eth` guard for WETH-priced pools** — `protocol.ts:122,126,132,136,174,178`. `polygon` and `celo` defensively return `null` if their price is unavailable, but `arbitrum`, `optimism`, and Base WETH-OLAS silently multiply by `prices.eth`. If `ethUsdPrice === '0'` you contribute `0` TVL/fees from three pools, totals still pass sanity, `stale=false`, and an artificially low POL is published with no fetchError flagged. Either mirror the `<= 0 ⇒ null` guard per-config, or short-circuit at top-of-function: if `prices.eth <= 0`, bail with `liquidity:ethereum` stale, since it's load-bearing.

3. **Docs/code drift in `ProtocolFeesInfo`** — `components/DataPage/ProtocolOwnedLiquidity.tsx:76-79` says L2 fees are converted via `2 × paired_token × price`. That's the TVL formula; fees are `(feePriced + feeOlas × reservePriced/reserveOlas) × price`. Please tighten so on-page docs don't contradict the implementation.

4. **`anyFeeMissing` → fallback path is implicit** — `protocol.ts:305,417`. If the ETH fee field is `null` and every L2 fee is out of bounds, `totalProtocolFeesUsd` stays at `0` (in-bounds), so you publish `0` with only `'liquidity:fees-partial'` in status. It only falls back because `createStaleStatus` treats any fetchError as `stale=true`, which `mergeWithFallback` then replaces. Works today, but relies on two indirections. Make it explicit: if `anyFeeMissing` then `feesValue = null`. Same intent, locally obvious.

5. **Per-pool fee breach doesn't push `fetchErrors`** (`protocol.ts:380-387`) while per-pool TVL breach does (`362-368`). Intentional fault isolation (POL shouldn't go stale for a fees-only bug), but undocumented. Add a one-line comment on that branch stating the intent — next person to touch this will \"fix\" it otherwise.

## Quality

6. **Six near-identical `computeTotalFeesUsd` closures** (`protocol.ts:99-156, 164-181`). Each unpacks `f0`/`f1` and calls `l2FeesToUsd` with the same plumbing. Lift to something like `{ pricedReserveKey: 'reserve0'|'reserve1', priceOf: (prices) => number, decimals?: number }` and derive both `computeTvlUsd` and `computeTotalFeesUsd`. Halves the file and makes a new chain a one-liner — the real long-term win.

7. **Serial `await getChainBlockNumber(chain)` in the L2 loop** (`protocol.ts:325`). Six sequential RPC round-trips on the critical path. Kick them off up front in parallel (`Promise.allSettled` alongside the subgraph fetches) and index into the result inside the loop.

8. **Asymmetric BigInt scaling in Base OLAS-USDC TVL** — `protocol.ts:164`: `(Number(BigInt(pool.reserve1)) / 1e6) * 2`. Safe today (USDC reserves ≪ 2⁵³) but inconsistent. Use `pairedFromReserve(pool.reserve1, 6)` for uniformity with the other pools.

9. **Hardcoded Base pool addresses** (`protocol.ts:162,172`). Consider moving to `common-util/constants.ts` next to other chain config. Also worth an explicit lowercase contract on the `.id` field — Balancer pool IDs sometimes come back checksummed depending on subgraph flavor.

10. **Magic sanity caps** (`protocol.ts:189-192`). Fine to ship as-is; a follow-up should source them from config so a future chain launch doesn't silently breach a globally-tuned cap.

## Nits

- `CLAUDE.md` addition is good; consider noting *when* to reach for the clamp pattern (cross-chain USD aggregation only, not single-chain calcs) so it's not over-applied.
- `.env.example` drops `DUNE_API_KEY`. Check that the Vercel project's prod/preview env also has it removed — orphan secrets linger.

## Verdict

Ship with (2) and (3) fixed in this PR. (4), (6), (7) are strong follow-ups; (1), (5), (8), (9) are quick cleanups worth folding in if you're already touching the file.

atepem:
## Additional review — POL/fees subgraph migration

Solid direction. A few more findings on top of what's already been flagged.

### Bugs / correctness

**1. Silent POL shrinkage when `computeTvlUsd` returns `null`** (`protocol.ts`, tvl-null branch). Stronger version of prior review point (2). The branch only sets `anyFeeMissing` and continues — it does not push to `fetchErrors`. So if `prices.matic <= 0` or `celoUsdPrice === '0'`, polygon/celo drop out of `totalPolUsd` silently, `polFetchErrors` stays empty, `stale=false`, and `mergeWithFallback` accepts the underreported value as fresh. The defensive `return null` guards in polygon/celo are actively *less* safe than the unguarded WETH paths here. Fix: push `source` to `fetchErrors` in the null branch, matching the TVL-breach branch above it.

**2. `originChain: 'base-weth'` key mismatch risk.** `bridgedBalances` is keyed off whatever `bridgedPOLHoldings[].originChain` the ETH subgraph emits. If it emits `'base'` for both Base LPs, the `'base-weth'` lookup falls back to `0n`, share=0, and WETH-OLAS contributes $0 POL with no flag raised. Confirm what the subgraph emits, or key bridged balances by pool address instead of a synthetic chain label.

**3. Missing `cumulativeFeesToken{0,1}` fields silently treated as 0.** `BigInt(pool.cumulativeFeesToken0 || '0')` swallows both the legitimate "no fees yet" case and a schema regression that drops the field. Distinguish `undefined` (→ `anyFeeMissing = true`) from `'0'` (→ legitimate zero).

**4. Base pools with unknown IDs are silently dropped.** `if (config) poolsToProcess.push(...)` skips any Base pool not in `BASE_POOL_CONFIG` with no log, no fetchError. At minimum `console.warn` on the miss so new pools don't ship POL-invisible.

**5. `anyFeeMissing` is a misnomer.** It's set on POL-TVL failures, pool-config misses, and subgraph failures — yet only influences `feesFetchErrors`, not `polFetchErrors`. Either rename (`anyPoolDataPartial`) or split into `polPartial` / `feesPartial` so POL status reflects its own partial-aggregation signal. Tangled with (1): once null-TVL pushes to `fetchErrors`, most `anyFeeMissing=true` assignments on those branches become redundant.

### Quality / consistency

**6. `feesFetchErrors` ternary redundancy.** `anyFeeMissing ? [...fetchErrors, 'liquidity:fees-partial'] : [...fetchErrors]` spreads twice. Collapse:

```ts
const feesFetchErrors = [...fetchErrors, ...(anyFeeMissing ? ['liquidity:fees-partial'] : [])];
```

**7. `source` is chain-level, not pool-level.** On Base, a per-pool breach pushes `liquidity:base` — you lose which of the two pools breached. Include `pool.id` (or a pool label) in the tag, not just in the log message.

**8. `Math.round(null) === 0` at the consumer** (`BondMetrics.tsx`). When the fetcher returns `value: null` (sanity breach, pre-merge), `Math.round(null)` coerces to `0` and the card renders `$0`. `mergeWithFallback` catches this on snapshot-read but not on raw-fetcher paths (dev, force-refresh, cold cache). Guard with `value == null ? null : Math.round(value)` so stale/empty rendering kicks in.

**9. `computeShare` can return > 1.** If `currentBalance` indexes later than `totalSupply` transiently, share exceeds 1 and overcounts POL. Low-impact; clamp with `Math.min(1, share)` and log.

**10. ETH fees null-vs-undefined double-duty.** `ethProtocolFeesRaw != null` intentionally catches both `null` and `undefined`. Fine, but worth a one-line comment so the `!= null` (vs `!==`) isn't read as a typo.

### Nits

- Feesvalue when everything fails: agreed with prior (4) — make `feesValue = null` on `anyFeeMissing` explicit, don't rely on the stale→fallback indirection.
- `liquidityL2Query` uses `first: 10` — fine, but add `orderBy: id, orderDirection: asc` so Base pool dispatch is deterministic across reindexes.
- `cumulativeFeesUsd` / `cumulativeExternalFeesUsd` typed but unread — one-line comment (`// fetched for schema parity`) if keeping.

### Verdict

Ship with prior-reviewer (2) and (3) addressed. Of the above, **(1), (2), (5)** are correctness issues worth folding into this PR since you're in the files already; **(3), (4), (7), (8)** are quick cleanups; the rest can wait for the follow-up refactor.

kupermind:
Thanks both — review-response commit pushed (`052bfaaa`). Breakdown:

## @Tanya-atatakai

- **#1 dead fields** — dropped `cumulativeFeesUsd` / `cumulativeExternalFeesUsd` from both the query and the type; left a comment on the type noting the field choice.
- **#2 missing eth-price guard** — added per-config `prices.eth <= 0` guards on arbitrum / optimism / base-weth (matches the polygon / celo style). Chose per-config over top-level bail so gnosis / celo / base-USDC don't go dark on an ETH-price failure. Commented the rationale next to `POOL_CONFIG_BY_CHAIN`.
- **#3 docs/code drift** — `ProtocolFeesInfo` now documents the actual formula: `(fee_priced + fee_olas × reserve_priced / reserve_olas) × price`, then × share.
- **#4 feesValue explicit** — done. `feesValue = feesPartial ? null : round(total)`. No more stale→fallback indirection.
- **#5 per-pool fee-breach comment** — added; also split `polPartial` / `feesPartial` so the semantic is enforced, not just documented (fees-only breach doesn't push to POL's fetchErrors).
- **#6 computeTotalFeesUsd dedup** — deferring. Not tiny, and the closures now each have independent guards (`prices.eth <= 0`, `parseFeesPair`, celo's `celoUsdPrice` check) that would need to be threaded through a declarative shape. Happy to do it in a follow-up once the config surface stops churning.
- **#7 parallel block lookups** — done. Kicked off in the same `Promise.allSettled` as the subgraph fetches; L2 loop now indexes into pre-resolved results.
- **#8 pairedFromReserve for USDC** — done; Base OLAS-USDC TVL now uses `pairedFromReserve(pool.reserve1, 6) * 2`.
- **#9 hoist Base addresses + lowercase `.id`** — done. `BASE_POOL_OLAS_USDC` / `BASE_POOL_WETH_OLAS` as module consts; dispatch site lowercases `pool.id` (already did, keeping it for safety).
- **#10 caps as config** — deferring; left them as named module consts with the pattern documented in CLAUDE.md. Worth sourcing from config when we have one more consumer of the pattern.
- **CLAUDE.md nit** — tightened to note the clamp is for *cross-chain USD aggregation*, not single-chain calcs.
- **DUNE_API_KEY in Vercel** — still on the post-merge to-do; called out in the PR description.

## @atepem

- **#1 silent POL shrinkage on null TVL** — fixed. `tvl === null` now pushes `${source}:${poolId}` to `fetchErrors` and sets both `polPartial` + `feesPartial`.
- **#2 `base-weth` key mismatch** — added a `console.warn` when a pool's `originChain` is absent from `bridgedBalances`. Keeps the (today-correct) chain-label scheme but makes the invariant self-announce if the subgraph ever changes.
- **#3 undefined vs '0'** — `parseFeesPair()` helper distinguishes `undefined` (→ null → fees partial) from `'0'` (→ legitimate no-fees-yet).
- **#4 unknown Base pool warn** — `console.warn` at the dispatch site for any Base pool id not in `BASE_POOL_CONFIG`.
- **#5 `anyFeeMissing` misnomer** — split into `polPartial` / `feesPartial`, each producing its own `*-partial` tag in the respective status.
- **#6 ternary redundancy** — cleaned up with the spread-conditional collapse.
- **#7 pool-level source tag** — per-pool errors now tagged `liquidity:${chain}:${poolId}`; chain-wide failures keep `liquidity:${chain}`.
- **#8 `Math.round(null)`** — `BondMetrics` now uses a `roundOrNull` helper so a sanity breach renders the stale state instead of `$0`.
- **#9 `computeShare > 1`** — clamped to `[0, 1]` with a warn on overshoot.
- **#10 `!= null` comment** — added; flagged as intentional (catches null and undefined).
- **Nit: `orderBy id asc`** — applied to `liquidityL2Query` for deterministic Base dispatch.

## Verification

- `yarn lint` / `tsc --noEmit` clean.
- Ground-truth check against `pol-aggregation.js --json`: totalPol ≈ \$2.48M, totalProtocolFees ≈ \$2.66M (values drifted slightly with ETH price since yesterday; structure matches).

rajat2502 (approved):
Solid fix with clean root-cause analysis. The BigInt-safe helpers correctly address the Number precision loss that caused the $1.14×10²⁰ bug, and the sanity-clamp is a good defensive layer.

Approving. A few non-blocking suggestions for a follow-up:

**Nits / follow-ups:**

1. **POL vs fees null-out asymmetry** — `feesValue` is nulled on any partiality, but `polValue` stays set when `polPartial` is true (only the final out-of-bounds check nulls it). The "explicit null is locally obvious" rationale applies equally to POL. If intentional (ETH-only POL still useful; partial L2 fees skew total), a one-line comment would help.

2. **Missed out-of-bounds tag on fees** — the `isSaneUsd(totalProtocolFeesUsd, MAX_TOTAL_FEES_USD)` check is behind `feesValue !== null`, so if `feesPartial` already nulled the value, the `liquidity:fees-out-of-bounds` tag never gets added. Observability-only, but consider splitting the tag push from the nullification.

3. **Unit tests for the new helpers** — strongly recommend adding tests for `l2FeesToUsd`, `pairedFromReserve`, `computeShare`, and the per-pool configs. They're pure functions, the decimals/reserve ordering is exactly what silently regressed before, and this would catch the next decimals bug at CI time rather than at Vercel-refresh time.

4. **PR description** — says "~200× headroom" for the $1B POL cap; actual is ~400× over $2.5M. Minor.

5. **Pool-id dispatch** — `pool.id.toLowerCase()` works today but would fall through to the "unknown pool" warning if a subgraph ever returned composite IDs (e.g. `${pool}-${block}`). A short comment noting the expected shape would help future-you.

6. **`l2Results[i * 2]` / `[i * 2 + 1]` indexing** after `flatMap` is correct but a bit fragile — a small helper to build `{ chain, subgraphPromise, blockPromise }` tuples before `Promise.allSettled` would read more clearly.

Good call on the `CLAUDE.md` addition and the `DUNE_API_KEY` cleanup with the explicit Vercel env delete action item.

Tanya-atatakai (approved):

DavidMinarsch (approved):

PR #508, opened 2026-05-22
ops: supply-chain hardening + clear Dependabot advisories

mohandast52:
Adopts the fleet-standard supply-chain pattern on top of the existing audit/lockfile CI gates, and clears the two open Dependabot alerts.

## Hardening (`ops`)
- **Audit wrapper** — `scripts/audit.mjs` (`yarn audit:prod`) replaces the inline audit step; handles the Yarn 1.x severity bitmask and suppresses only IDs in `.supply-chain/audit-allowlist.json` (currently **empty**), warning on expired review dates.
- **Install-hook diff gate** — `scripts/audit-install-hooks.mjs` + `.supply-chain/install-hooks.allowlist`, wired into `all-checks-passed`; catches new transitive (pre/post)install scripts that `yarn audit` never sees.
- **gitleaks** secret-scan workflow (SHA-pinned, separate workflow).
- **`.yarnrc`** skips install scripts by default. Verified a clean `node_modules` reinstall still lints + builds — `sharp`/`unrs-resolver`/`bufferutil` ship prebuilt platform binaries, so no lifecycle script is needed at build time.
- **`.nvmrc`** `22.22.3`, read by CI via `node-version-file`.
- **`.github/CODEOWNERS`** routes everything to `@Tanya-atatakai @atepem`.
- **Carets stripped** from `package.json` `resolutions`.
- **`SUPPLY-CHAIN-SECURITY.md`** refreshed (audit wrapper, install-hook gate, secret scanning, `.yarnrc` policy).

## Dependabot fix (`fix`)
Closes both open alerts via exact resolutions (verified `lint`/`build` pass with the forced `postcss` override into Next):
- `postcss@8.5.10` — Next bundled `8.4.31` (GHSA-qx2v-qp2m-jg93, medium)
- `@babel/plugin-transform-modules-systemjs@7.29.4` — `7.27.1` via `@babel/preset-env` (GHSA-fv7c-fp4j-7gwp, high, dev-scope)

## Verification
`yarn lint`, `yarn build`, `yarn audit:prod` (0 high/critical), `yarn audit:install-hooks` all pass on Node 22. `yarn.lock` moved only by the deliberate resolution bumps.

## Follow-ups before marking ready
- [ ] Confirm `@Tanya-atatakai`/`@atepem` have **write access** (CODEOWNERS ignores owners without it) and enable **"Require review from Code Owners"** in branch protection.
- [ ] Add **gitleaks as its own required status check** (separate workflow → not covered by `all-checks-passed`).
- [ ] **Sandbox-test each gate's failure mode** (vuln dep → audit fails; new postinstall → install-hooks fails; fake secret → gitleaks fails).
- [ ] Dependabot alerts close automatically once this merges to `main`.


Tanya-atatakai (changes requested):
Reviewed against the fleet reference in `autonolas-frontend-mono/scripts/` (`audit.mjs` + `audit-install-hooks.mjs`) — which is the upstream the PR's own comments asked to reconcile against once reachable. The PR is substantive and mostly well-done, but the two scripts have real coverage and correctness gaps vs. the reference. Leaving as request-changes because findings 1–4 are not just style differences — they're things that make the gates silently weaker than intended.

## Worth applauding

- `gitleaks` pinned by version **+ sha256 checksum**, run via the open-source CLI (license-free on org repos). Better posture than just SHA-pinning the wrapper.
- `.yarnrc --install.ignore-scripts true` is a defense-in-depth layer that even the autonolas-frontend-mono monorepo doesn't have yet.
- Carets stripped from `resolutions` — consistent with the exact-version policy.
- `_keep_when_empty` annotation in `audit-allowlist.json` is a thoughtful note for future archaeologists.
- `audit:install-hooks:update` preserves human-written justification lines and tags genuinely-new entries `TODO: justify`. That UX is *better* than the upstream reference (which clobbers).
- CODEOWNERS routes the supply-chain surface cleanly. The PR description correctly flags branch-protection follow-ups.
- Dependabot fixes (`postcss 8.4.31 → 8.5.10`, `@babel/plugin-transform-modules-systemjs → 7.29.4`) look targeted; `yarn.lock` diff is scoped to those bumps and the expected transitive churn.

## Findings (fleet-parity divergences)

**1. `scripts/audit-install-hooks.mjs` doesn't recurse into nested `node_modules/` — coverage gap**

The PR's `packageDirs()` (L30-43) only enumerates `node_modules/<pkg>` and `node_modules/@scope/<pkg>`. The reference in `autonolas-frontend-mono/scripts/audit-install-hooks.mjs:70-95` recursively walks nested `node_modules/` with `MAX_DEPTH=20` and a depth guard. With Yarn 1.x hoisting most deps land at the top level, but version conflicts produce nested copies — a malicious install hook in a nested dep is invisible to the gate as written. Port the recursive walk.

**2. `scripts/audit-install-hooks.mjs` has no allowlist-drift detection**

The PR's gate fails on `tree ∖ allowlist` (new hook in tree) but **silently ignores** `allowlist ∖ tree` (stale entry — dep was removed). The reference fails on both directions. Without bidirectional drift detection, dead entries accumulate and erode the meaning of "every entry was reviewed."

**3. `scripts/audit.mjs` doesn't validate the allowlist schema**

`loadAllowlist()` (L23-32) parses JSON but never validates the fields. The reference (`autonolas-frontend-mono/scripts/audit.mjs:30-41`) enforces at load time: `id` is a number, `reason` is a non-empty string, `review` matches `^\d{4}-\d{2}-\d{2}$`, exits code 2 on malformed entries. Without this, a typo'd entry (empty `reason`, missing `review`, wrong-type `id`) silently suppresses real advisories. The schema is *documented* in `_fields` but never enforced. Port the validation.

**4. `scripts/audit.mjs` doesn't warn on no-longer-suppressing allowlist entries**

The reference (L128-133, L152-156) tracks which allowlist entries are *actually* suppressing a current high/critical advisory and emits a `::warning::` for ones that aren't (upstream fix landed → entry is doing no work). Without this signal, dead allowlist entries accumulate silently and the "every entry was justified at some point" invariant decays into "every entry was justified at some point, but most of them are now lies."

**5. Trivial-hook fast path is intentionally omitted — surface this as a design decision**

The reference (`autonolas-frontend-mono/scripts/audit-install-hooks.mjs:47-52`) treats `echo "..."`, `true`, `:`, and `exit 0` as non-hooks via a `TRIVIAL` regex set. The echo pattern uses a negative-lookahead to reject shell-metachar prefixes (`echo "ok" && curl …`) so an attacker can't slip past by prefixing `echo`. The PR's author explicitly chose not to include this ("intentionally does NOT auto-allow 'trivial' hooks — every hook must be explicitly allowlisted"). That's a defensible design call — the explicit allowlist is more conservative — but it's a real fleet divergence, and benign trivial-hook packages will need allowlist entries forever. Worth surfacing as a deliberate design decision in `SUPPLY-CHAIN-SECURITY.md` rather than an unstated divergence in a code comment, so future readers know which way to go when the two repos drift further.

**6. `.gitleaks.toml` allowlist suppresses whole lines — minor risk**

`regexTarget = "line"` with `regexes = ['''0x[a-fA-F0-9]{40}''']` suppresses the *entire matching line*. Practically fine because secrets and EVM addresses don't co-occur in this codebase. But if they ever did (`SLACK_TOKEN=xoxb-... # for monitoring 0xabc...`), the secret would be hidden. A tighter form is `regexTarget = "match"` with the same regex. Low-priority — flagging because it's a documented gitleaks footgun.

## Nits

- `scripts/audit.mjs:32` has unreachable `return { entries: [] };` after `process.exit(2)`.
- `scripts/audit.mjs:38` uses `String(e.id)` for allowlist matching — loose vs. the reference's strict numeric schema. Works either way as long as JSON `id` stays consistent.
- The install-hook walker doesn't explicitly handle symlinks. The reference uses `Dirent.isDirectory()` (false on symlinks) to skip. Symlink cycles in `node_modules` are rare in Yarn 1.x trees, but worth a defensive line.

## Suggested path forward

Copy the two scripts verbatim from `valory-xyz/autonolas-frontend-mono/scripts/` (they're MIT-compatible, same org/author) and re-justify the four existing `install-hooks.allowlist` entries. That closes findings 1–4 in one shot. Findings 5 and 6 are design calls — pick a direction and document it. The PR's other improvements (`.yarnrc`, the better `--update` UX, gitleaks CLI pinning, CODEOWNERS routing, `_keep_when_empty` annotation) are worth porting back into the monorepo as a follow-up.

Tanya-atatakai (approved):
Re-reviewed `4dc92a4e fix(ci): align audit scripts with autonolas-frontend-mono reference`. All 6 findings + the 3 nits are cleanly resolved.

### Findings → fixes (verified)

1. **Nested `node_modules/` recursion** — `walkPackageJsons` ported with `MAX_DEPTH = 20`, scoped-package handling, nested recursion, and `Dirent.isDirectory()` symlink skip. Coverage gap closed.
2. **Bidirectional drift detection** — `unexpected` (tree ∖ allowlist) AND `missing` (allowlist ∖ tree) both reported with separate error messages; gate fails if either set is non-empty.
3. **Schema validation in `audit.mjs`** — `loadAllowlist` now validates each entry at load: `typeof id === 'number'`, non-empty trimmed `reason`, `review` matching `^\d{4}-\d{2}-\d{2}$`. Exits 2 on malformed entries. No more silent acceptance.
4. **Stale-entry warning** — Verified in the deployed file: `suppressedIds` is built from currently-matching advisories, then a `for (const entry of allowlist.entries)` loop emits `::warning::` for any entry not in that set. The comment explaining why we compare against the *suppressed* set (not the full advisory list) is a nice touch — captures the "severity dropped below high/critical" edge case.
5. **Trivial-hook fast path** — Added with the hardened echo lookahead `/^(?!.*[&|;\`$()<>])echo(\s|$)/` that rejects shell-metachar chaining. Documented in SUPPLY-CHAIN-SECURITY.md as a deliberate parity choice rather than an unstated divergence.
6. **`.gitleaks.toml` line-level allowlist** — Flipped to `regexTarget = "match"` with an explanatory comment; an EVM address can no longer mask a real secret on the same line.

### Nits also resolved

- Unreachable `return { entries: [] };` after `process.exit(2)` — gone.
- `String(id)` loose matching — replaced with strict numeric ID matching.
- Symlink handling — now explicit via `Dirent.isDirectory()`.

### Preserved local improvements

Good call keeping these as deliberate local divergences, called out in the script header:

- `--update` preserves human justification comments and tags only genuinely-new packages with `TODO: justify` (the upstream reference clobbers).
- `_keep_when_empty` annotation in `audit-allowlist.json`.
- `.yarnrc --install.ignore-scripts true` defense-in-depth (the monorepo lacks this).
- gitleaks pinned by version + sha256.

### Docs aligned

`SUPPLY-CHAIN-SECURITY.md` was updated to match: schema validation, drift detection, recursive scan, trivial-hook handling — all explicitly documented, plus the "Mirrors `valory-xyz/autonolas-frontend-mono/scripts/...`" pointer makes the parity intent durable.

Solid fleet adoption. Worth porting the four preserved improvements back into autonolas-frontend-mono as a follow-up (`.yarnrc`, the `--update` UX, gitleaks-cli pattern, `_keep_when_empty` annotation).

rajat2502 (approved):

PR #513, opened 2026-06-02
feat(predict): (BDMARK-2173) add Brier score to Omenstrat performance card

atepem:
## Proposed Changes

Consume the Brier-score accumulators shipped in predict-omen (autonolas-subgraph PR #95). The Omenstrat economy "Performance" card now shows ROI / Prediction Accuracy / Brier Score (replacing OLAS Staking APR), with the 7D/30D/90D/Max time-range tabs windowing the Brier value.

- queries.ts: lean getOmenDailyBrierStatsQuery (date, brierSum, brierCount, _meta)
- omenstrat-brier.ts: paginate dailyProfitStatistics; per window compute sum(brierSum)/sum(brierCount) -> WindowedMetric<number|null>, with lag/stale handling
- predict/index.ts: wire brierScore into PredictMetricsData.omenstrat snapshot
- PlatformActivitySection: activeWindow state, enable tabs on Omenstrat (disabled on Polystrat until its subgraph ships Brier), Brier metric + "lower is better" popover
- Activity.tsx: map brierScore value/status through processPredictMetrics

Polystrat keeps APR for now (predict-polymarket Brier is a future PR). Field is additive to the snapshot, so no METRICS_PREFIX bump is needed.

## Screenshots/Recordings

image

## Things to keep in mind

### DIFF
--- a/common-util/api/predict/index.ts
+++ b/common-util/api/predict/index.ts
@@ -17,6 +17,7 @@ import {
 import { MetricWithStatus, WithMeta } from 'common-util/graphql/types';
 import { getMaxApr } from 'common-util/olasApr';
 import { getMidnightUtcTimestampDaysAgo } from 'common-util/time';
+import { fetchOmenstratBrier, WindowedMetric } from './omenstrat-brier';
 import { fetchOmenstratRoi } from './omenstrat-roi';
 import { fetchOmenstratSuccessRate } from './omenstrat-success-rate';
 import { fetchPolystratRoi } from './polystrat-roi';
@@ -211,6 +212,8 @@ const fetchPolystratOlasApr = async (): Promise<MetricWithStatus<string | null>>
   });
 };
 
+export type { WindowKey, WindowedMetric } from './omenstrat-brier';
+
 export type PredictMetricsData = {
   // Omenstrat metrics
   omenstrat: {
@@ -220,6 +223,8 @@ export type PredictMetricsData = {
     partialRoi: MetricWithStatus<number | null>;
     finalRoi: MetricWithStatus<number | null>;
     successRate: MetricWithStatus<string | null>;
+    // Windowed mean Brier score (predict-omen only). Lower is better; ~0.25 = a 50/50 guess.
+    brierScore: MetricWithStatus<WindowedMetric<number | null>>;
   };
 
   // Polystrat metrics
@@ -246,6 +251,7 @@ export const fetchAllPredictMetrics = async (): Promise<PredictMetricsSnapshot |
       omenstratTxsResult,
       omenstratRoiResult,
       omenstratSuccessRateResult,
+      omenstratBrierResult,
       polystratAprResult,
       polystratTxsResult,
       polystratRoiResult,
@@ -258,6 +264,7 @@ export const fetchAllPredictMetrics = async (): Promise<PredictMetricsSnapshot |
       fetchPredictTxsByAgentType(),
       fetchOmenstratRoi(),
       fetchOmenstratSuccessRate(),
+      fetchOmenstratBrier(),
       // Polystrat
       fetchPolystratOlasApr(),
       fetchPolystratTxsByAgentType(),
@@ -309,6 +316,13 @@ export const fetchAllPredictMetrics = async (): Promise<PredictMetricsSnapshot |
           omenstratSuccessRateResult.status === 'fulfilled'
             ? omenstratSuccessRateResult.value
             : { value: null, status: getFetchErrorAndCreateStaleStatus('omenstrat:successRate') },
+        brierScore:
+          omenstratBrierResult.status === 'fulfilled'
+            ? omenstratBrierResult.value
+            : {
+                value: { '7d': null, '30d': null, '90d': null, max: null },
+                status: getFetchErrorAndCreateStaleStatus('omenstrat:brier'),
+              },
       },
 
       polystrat: {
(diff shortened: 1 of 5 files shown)


mohandast52 (commented):
Reviewed the full diff against the surrounding patterns. Overall this is clean and well-commented — nice work. A few notes, one of which I'd like to confirm before this ships.

### 🔴 Main concern: all-history pagination on the hourly cron

`fetchOmenstratBrier()` paginates `dailyProfitStatistics` from `date_gte: 0` (all time) on **every** run of `/api/refresh-metrics/predict`, which is scheduled **hourly** (`vercel.json`: `0 * * * *`).

That entity is per-`(traderAgent, day)`, not global-per-day — `roi-distribution.ts` groups multiple rows per day and reads `stat.traderAgent.id` (`roi-distribution.ts:457-480`). So the unfiltered query returns `agents × active-days` rows across the whole history.

Two problems follow:

1. **Skip cap.** Skip-based pagination through The Graph gateway/Studio is capped at `skip ≤ 5000`. With `LIMIT = 1000`, once the dataset exceeds ~6000 rows the loop hits `skip = 6000`, the request throws, the `catch` returns `emptyWindows()` + `fetchError`, and the metric is *perpetually* stale (it never populates). Worth confirming whether this subgraph is served via the gateway/Studio (cap enforced) or a self-hosted graph-node (cap may be absent).
2. **Unbounded growth.** Even if today's row count is under the cap, `date_gte: 0` never advances, so this query grows forever and re-scans the entire history every hour.

The sibling `roi-distribution.ts` already navigates exactly this: it caps at `MAX_DAYS_PER_RUN = 30` per run and maintains an incremental `allTimeAgents` blob for the "max" aggregate — strong evidence the dataset is large enough that a single all-history pass is a problem.

Suggested directions:
- Bound the windowed query to `date_gte: cutoff90d` — that covers 7D/30D/90D and caps the row count to ~90 days.
- For **Max**, read a global all-time accumulator from the subgraph if predict-omen exposes one (the PR description mentions "Brier-score accumulators" — is there a `global { brierSum brierCount }` singleton?). That makes Max O(1) instead of a full history scan.

If you've already verified the all-time row count stays comfortably under the cap, a comment to that effect would be enough for now — but the unbounded-growth angle still makes me lean toward windowing the query.

### 🟡 Minor

- **No `/data` provenance anchor for Brier.** Every other performance metric links to `/data#<platform>-predict-*`; `brierItem` has no `href`. Given the site's "Data Verification" page documents metric provenance, consider a follow-up to add a Brier section + anchor.
- **`_meta` checked only on the first page** (`metaChecked`). Fine and efficient — just flagging that `roi.ts`'s `fetchMechRequests` re-checks every page, so the two pagination loops differ slightly. Not a blocker.
- **Duplicated empty-windows literal.** `index.ts`'s fetch-error fallback inlines `{ '7d': null, '30d': null, '90d': null, max: null }`, which `emptyWindows()` already produces — could export and reuse it. Trivial.
- **Window boundary is inclusive** (`date >= cutoff7d` includes the bucket from exactly 7 days ago → 8 daily buckets). Assuming that's the intended "last 7 days" semantics; just confirming it's deliberate.

Holding off on approval pending the pagination question — happy to approve once that's clarified or addressed.

mohandast52 (dismissed):
Re-reviewed after the incremental rework (`c9aac46`, `bd87aaa`) — this fully addresses the pagination blocker. 👍

The all-history rescan is replaced by a persisted accumulator blob (`predict-brier/omenstrat`) that advances a bounded amount per hourly run: a ~10-day head refresh (with gap-bridging if a cron outage left `coveredTo` behind) plus a one-time 30-day backward backfill to genesis. Every `dailyProfitStatistics` query is now bounded to a ≤30-day range, so none approach The Graph's skip cap. Verified along the way:

- `OMEN_GENESIS_DAY` matches `OMEN_GENESIS_TS` in `roi-distribution.ts` and is a true UTC midnight.
- `fetchAllPredictMetrics` (which now writes a blob) runs only in the hourly cron, not at ISR/build.
- Buckets are keyed by day and **overwritten**, so reprocessing the trailing window can't double-count.
- The `backfilling` fetchError marks the metric stale during catch-up, so `mergeWithFallback` serves the last fully-computed value — and the accumulator is saved with `overwrite: true`, which is the right call (a merge would resurrect deleted buckets).
- Bonus: `emptyWindows` is now shared, and the window math excludes today, fixing the off-by-one from the first round.

Approving. A few **non-blocking** notes for whenever convenient:

1. **Frozen historical buckets vs. late re-answers.** Buckets older than `TRAIL_DAYS` (10) are never refreshed. The "Brier moves to the new settlement day on a re-answer" comment makes overwriting the trailing window correct *only if* re-answers land within ~10 days. If predict-omen's arbitration/dispute window can push a re-answer past 10 days, the old day's (now-reduced) bucket stays inflated while the new day is also counted → slight Max/90D drift. Worth confirming the max re-answer lag is comfortably under `TRAIL_DAYS`.
2. **Transient accumulator-read failure restarts backfill.** If `getSnapshot` for the accumulator fails mid-backfill, `existing` is null and `buckets` rebuilds from `{}`, discarding backfill progress (it stays stale and re-converges, so no incorrect value is published — just a slower Max during catch-up). Fine as-is; flagging in case you'd rather retry the read.
3. **Slashed category** (`predict-brier/omenstrat`) yields a nested blob pathname — first time the codebase does this. `list({prefix})` + exact `pathname` match should handle it, but worth eyeballing the first cron run to confirm the round-trip.
4. **Carried over from round 1:** Brier still has no `/data#...` provenance anchor like the sibling metrics — a follow-up to add a Data Verification section would be nice.

Nice work on the rework.

rajat2502 (approved):

DavidMinarsch (approved):

PR #515, opened 2026-06-05
feat: Agent Economy Explorer page with DAA, Transactions, Accuracy & ROI heatmaps

atepem:
## Summary

Adds the **Agent Economy Explorer** at `/explorer` — a visual, time-aware view of
Olas agent-economy activity. Each metric tile is clickable and drives a
GitHub-style calendar heatmap with hover tooltips. Covers the **Predict economy
(Omenstrat)** today; the economy selector is built to extend to others.

Four metrics:
| Metric | Source | Coverage |
|---|---|---|
| Daily Active Agents | registry subgraph (`dailyAgentPerformances`) | full history (~2.5 yr) |
| Transactions | registry subgraph (same query) | full history |
| Prediction Accuracy | predict-agents subgraph (resolved bets → daily win-rate) | backfilled window |
| Return on Investment | predict-agents subgraph (`dailyProfitStatistics`) | backfilled window |

## How it works

- **Snapshot pipeline** (matches the rest of the site): a daily cron
  `refresh-metrics/explorer` writes a Vercel Blob snapshot; the page reads it in
  `getStaticProps` (ISR), so **visitors/ISR never hit the subgraphs** — only the
  cron does.
- **DAA/Transactions** are cheap (one pre-aggregated row/day) → full-refetched each run.
- **Accuracy/ROI** have no daily aggregate upstream, so they're computed from raw
  records (bets / per-agent rows). To avoid re-paging deep history every run, they use
  **backfill-once + daily window-merge**: the cron fetches only a recent window and
  merges it into the stored series (recent days overwrite to self-heal late-settling
  markets, new days append, older history is kept). Accuracy paginates by **cursor**
  (`timestamp_lt`) so a deep backfill stays O(1)/page; failures are isolated and keep
  prior history.
- **Heatmap**: memoized cell grid with precomputed colors; sequential purple ramp for
  counts/accuracy, **diverging red(loss)/green(profit)** for ROI, with a scale-aware legend.

## Video

https://github.com/user-attachments/assets/f5eb0abb-cdac-4b40-aee3-1904e4a6095e

## Deployment — required one-time backfill ⚠️

The production `explorer` blob is separate and starts empty. After deploy, trigger a
**one-time backfill** so Accuracy/ROI have history (DAA/Tx fill themselves):

`GET /api/refresh-metrics/explorer?accuracyPages=400&roiDays=365`

(Fits the 300s function limit: ~5 months Accuracy, ~12 months ROI.) The daily cron
(`0 6 * * *`) then maintains it. Until the backfill runs, Accuracy/ROI tiles show `--`.

## Notes / caveats

- ROI uses **partial ROI** = `profit / (payout − profit)` (the cost basis this subgraph
  version exposes), not the mech-fee-inclusive figure on the Predict page.
- Accuracy/ROI are bounded windows by design (raw-record volume); the *optimal* long-term
  fix is a daily-aggregate entity in the predict subgraph, which would make them as cheap
  as DAA. Tracked as a follow-up.
- New route is `/explorer`; the pre-existing `/explore` → `/agents` redirect is unrelated.

## Testing

- `yarn lint` + `tsc --noEmit` clean.
- Verified end-to-end locally: backfill (Accuracy 205d, ROI 362d), daily-cron simulation
  preserves history (~7s), metric switching, diverging ROI colors, hover tooltips.


mohandast52 (commented):
Reviewed the full diff end-to-end (data pipeline, cron, snapshot wiring, page, and the heatmap component). This is a polished, well-architected feature — it follows the site's snapshot→Blob→ISR pattern faithfully (visitors never hit subgraphs), the backfill-once + window-merge design for Accuracy/ROI mirrors the approach we just landed in #513, and the heatmap component is genuinely nice work (memoized ~1.3k-cell grid kept off the scroll/hover hot path, timer/listener/ResizeObserver cleanup, `prefers-reduced-motion`, quantile sequential + diverging color scales, wheel-pan that releases the page at the track ends). A11y on the selectors (role/aria-selected/aria-disabled/titles) is a nice touch.

Verified along the way: registry `agentIds` is `[14, 25]` (2 agents → ~2 pages, `MAX_PAGES=8` is a safe backstop), `txCount` is an established field, `mergeWithFallback` handles the whole-value registry-failure fallback correctly (registry drives staleness; a flaky Accuracy/ROI run keeps prior history via `mergeSeries`), and `/explore`→`/agents` is an exact-match redirect so `/explorer` is unaffected.

One issue I'd like addressed before the documented backfill is relied on, plus a few minor notes.

### 🔴 ROI uses skip-pagination across a wide window — the `roiDays=365` backfill won't actually backfill a year

`getExplorerDailyProfitStatsQuery` pages with `skip: page * ROI_PAGE_SIZE` (`fetchOmenstratDailyRoi` in `common-util/api/explorer.ts`). But `dailyProfitStatistics` is per-(agent, day) — the code itself estimates `ROI_ROWS_PER_DAY_EST = 300`. So a `roiDays=365` backfill needs to page to `skip ≈ 110k`, and The Graph gateway caps `skip` at ~5000–6000 — the *exact* constraint your Accuracy query comment already calls out ("skip pagination is O(skip) on The Graph and the gateway rejects deep skips").

Concretely: the deep backfill throws around page 6 (`skip=6000`), the `catch` breaks the loop keeping ~6000 rows ≈ the most recent **~20 days**, not 365. So the deployment step in the PR description —

> `GET /api/refresh-metrics/explorer?accuracyPages=400&roiDays=365` … ~12 months ROI

— silently yields ~20 days of ROI history instead of a year. (ROI then grows ~1 day per daily cron run going forward, so it self-heals over weeks/months, but the one-time historical backfill doesn't do what the docs say.) The daily cron itself (`roiDays=30`) is fine in practice — it still captures the recent ~20 days it needs to self-heal.

This is the same class of issue #513 hit, and you've already solved it for Accuracy in this very PR — ROI just needs the same treatment. Suggested fix: page ROI by a `date_lt` cursor (or per-day like `roi-distribution.ts`) instead of `skip`. As a bonus, the "if this page cap ever truncates … it drops the OLDEST days" comment in `fetchOmenstratDailyRoi` only holds for a clean truncation — today it's a thrown request caught mid-stream, so making it cursor-based also makes that comment true.

### 🟡 Minor

1. **Accuracy cursor can drop same-second bets at page boundaries.** `cursor = pageRows[last].timestamp; … timestamp_lt: cursor` excludes any not-yet-fetched bets sharing that exact second. At ~1000 bets/page in a busy market, boundary collisions are plausible, dropping a few bets per page and slightly skewing that day's denominator. Using `timestamp_lte` + dedup on a unique id (or an id cursor) would be exact. Low impact.
2. **Deep-backfill runtime.** `accuracyPages=400` is up to 400 sequential paged requests (× retries) in a single 300s function; if it overruns, the whole invocation dies and nothing is saved. You note it fits — might be worth documenting a safe ceiling / splitting very deep backfills.
3. **Stale comment.** `styles/globals.css` references flipping `USE_PROXIMITY_HOVER` in `DaaCalendarHeatmap.tsx`, but that flag isn't in the committed component (proximity hover was removed). Tidy the comment.
4. **Nits:** `vercel.json` has no trailing newline; `revalidate: 3600` re-reads the daily blob ~24×/day (harmless, and the comment explains the intent).

Happy to approve once ROI is cursor-paged (or the `roiDays` backfill claim is adjusted to match what skip-pagination can actually reach). Really nice feature overall.

rajat2502 (commented):
## Review — Agent Economy Explorer

Reviewed against `main`, 16 files (+1923/−1). New `/explorer` route; only `snapshot-storage.ts`, `queries.ts`, `vercel.json`, `globals.css` are modified (all additive).

### Intent

Adds a GitHub-style calendar-heatmap explorer for the Predict (Omenstrat) economy with four clickable metric tiles: DAA + Transactions (cheap, full-refetch from the registry subgraph) and Accuracy + ROI (derived from raw predict-agents records via backfill-once + daily window-merge). Rides the existing snapshot+cron pattern — a daily cron writes a Vercel Blob, the page reads it in `getStaticProps`, so visitors/ISR never touch subgraphs. The heatmap component is the bulk of the diff (purple sequential ramp for counts/%, diverging red/green for signed ROI).

### 🔴 Blocking

**`components/ExplorerPage/index.tsx` — `activeYear` not reset on metric change.**
Select a year present in DAA (e.g. 2023), switch to ROI/Accuracy (bounded recent window without 2023) → every cell satisfies `cell.year !== highlightYear` and renders as a dashed empty outline. The whole heatmap blanks. Worse: the stale year's tab isn't in the new metric's `years`, so the user can't click it to un-toggle.
**Fix:** reset `setActiveYear(null)` when `activeMetric` changes, or clamp before passing down: `highlightYear={years.includes(activeYear) ? activeYear : null}`.

### Security

Reviewed input boundaries. The new GraphQL queries interpolate values into `gql` strings, but every interpolated value is numeric and internally derived (`Number()`-coerced cursors/timestamps, constants); `accuracyPages`/`roiDays` pass through `toInt()` (positive-int only) and only drive loop counts — never reach the query as strings. No injection sink. Matches the repo's existing query-builder convention.

One non-blocking note: `pages/api/refresh-metrics/explorer.ts` is unauthenticated (consistent with the other `refresh-metrics/*` endpoints — none use `CRON_SECRET`), but it newly accepts caller-controlled `accuracyPages`/`roiDays` that amplify subgraph load. `accuracyPages` is clamped (`ACCURACY_MAX_PAGES=800`); `roiDays` has no upper clamp. Consider clamping `roiDays` and/or gating these write endpoints behind `CRON_SECRET` as a follow-up for the whole family.

### Regression risks

- **`snapshot-storage.ts` union change is additive** — existing categories unaffected. Verified the failure path: a null `registry.value` → `newValueIsInvalid` in `mergeWithFallback` → prior blob (incl. accuracy/roi) is preserved. A flaky registry run won't blank the page.
- **Snapshot is read twice per run** (endpoint for `previous`, then inside `saveSnapshot`'s merge). Accuracy/ROI are already merged via `mergeSeries` before `mergeWithFallback` runs, so no double-merge corruption. Safe, minor inefficiency.

### Nits

- `common-util/api/explorer.ts` — accuracy cursor paging uses `timestamp_lt = last.timestamp` (non-unique key); if >`PAGE_SIZE` bets share the boundary second, the overflow is dropped on the next page. Low impact (a day's win-rate is a sample), worth a comment.
- `styles/globals.css` — `.scrollbar-subtle` is defined but unused anywhere (the heatmap uses `.heatmap-scroll`). Drop it or confirm intended for reuse.
- `styles/globals.css` — comment references `USE_PROXIMITY_HOVER` in `DaaCalendarHeatmap.tsx`, but that symbol doesn't exist (stale).
- `components/ExplorerPage/index.tsx` — headline Accuracy/ROI are unweighted means of daily values (a 12-bet day weighs the same as a 2000-bet day). Documented in the PR; fine for POC.

### Open questions

- **Backfill is a manual one-time step** (`?accuracyPages=400&roiDays=365`) and the prod blob starts empty — until it runs, Accuracy/ROI tiles show `--`. Is there a deploy runbook/owner for triggering it, or should it fold into the deploy?
- **Economy switching is inert** — `activeEconomy` state exists but only Predict has data (Babydegen/Mech disabled). Confirmed intentional (POC), just flagging the selector currently changes nothing.

rajat2502 (commented):
## Re-review — all points addressed ✅

Re-checked against `bc252480`.

- **🔴 Year filter (blocking)** — fixed. `handleMetric` resets `setActiveYear(null)` on metric switch; the blank-heatmap path is closed.
- **ROI deep-skip / unbounded `roiDays`** — fixed properly. Reworked from `skip`-paging the whole window to **date-range chunking** (7-day chunks, newest→oldest). Traced the boundaries: chunks are contiguous and non-overlapping (`chunkLte = chunkGte − 1 day`), each day counted exactly once, loop terminates at `chunkLte < oldest`, and a chunk failure keeps the most-recent chunks. `roiDays` now clamped to `ROI_MAX_WINDOW_DAYS = 800`. Old `ROI_ROWS_PER_DAY_EST` removed with no dangling refs.
- **Nits** — `.scrollbar-subtle` removed, stale `USE_PROXIMITY_HOVER` comment removed, accuracy cursor-boundary gap now has an explicit acknowledging comment.
- **Bonus** — relabeling to *Total Transactions / Avg Prediction Accuracy / Avg Trading ROI* makes the headline aggregation honest in the UI; nicely covers the unweighted-mean note.

No new issues introduced by the follow-ups. The two open questions (manual prod backfill ownership, inert economy selector) remain non-blocking. LGTM from my side.

rajat2502 (dismissed):

mohandast52 (dismissed):
Re-reviewed after the latest commits (`77dae7aa`, `5534b12b`, `bc252480`) — the ROI blocker is properly fixed and every other note is addressed. 👍

**ROI pagination** now pages the window in **non-overlapping 7-day date-range chunks** (newest→oldest), resetting `skip` per chunk so each chunk's max skip (~3k for 7d × ~500 rows/day) stays well under the gateway cap. I checked the boundaries: chunks are contiguous with no overlap and no gap (all aligned to midnight `date` values), so each date's per-agent rows are counted exactly once; the loop terminates cleanly (clamped to `oldest`); and a chunk failure keeps the most-recent chunks rather than just the oldest. So a `roiDays=365` backfill now genuinely reaches a year of history — and `ROI_MAX_WINDOW_DAYS=800` caps the unauthenticated endpoint against unbounded load. 

Also addressed:
- Stale `USE_PROXIMITY_HOVER` CSS comment removed (and the unused `.scrollbar-subtle` block dropped).
- Backfill 300s-limit caveat documented in the endpoint; accuracy's same-second `timestamp_lt` boundary tradeoff now documented as a conscious choice (fine — it's a daily-sample win-rate).
- `vercel.json` trailing newline.
- Bonus: clearer metric labels (Total Transactions / Avg Prediction Accuracy / Avg Trading ROI) that match the headline value semantics, and **resetting the year filter on metric switch** — a real UX fix (switching to a metric that lacks the selected year would otherwise dim every cell into a stuck-blank heatmap). Nice catch.

Approving. One **non-blocking** forward-looking note: the 7-day chunk gives ~1.7× headroom over the ~500 rows/day estimate (a chunk only fails if rows/day exceeds ~850). If per-day ROI row volume grows past that, drop `ROI_CHUNK_DAYS`. Not a concern today.

Great feature — thanks for the thorough iteration.

rajat2502 (dismissed):

mohandast52 (approved):

rajat2502 (approved):

PR #521, opened 2026-06-11
feat(predict): [BDMARK-2179] window ROI and accuracy

atepem:
## Problem

On `/agent-economies/predict`, the Performance card's 7D/30D/90D/Max tabs only fed the **Brier score**, so clicking them appeared to do nothing — ROI, APR, and Accuracy were scalar values that ignored the window, and on Polystrat (no Brier) every tab but Max was disabled.

## Change

- **Accuracy** is now windowed for both platforms — a Brier-style daily accumulator (`predict-accuracy/{omenstrat,polystrat}`) counting correct/total settled bets bucketed by placement day.
- **ROI** (partial + final) is now windowed — `windowed-roi.ts` combines per-window prediction net-gain/costs from the existing `roi-distribution` snapshot with windowed staking rewards (new `predict-staking-rewards/*` accumulator over `serviceRewardsHistories`) valued at the current OLAS price.
- **APR** stays an un-windowed current rate (it's a configured staking rate, no time series).
- The Performance card is split into a **"Current"** group (APR) and a **"By time range"** group (ROI, Accuracy, Brier) so the tabs visibly govern only the windowed metrics.

### Notable correctness fix
The Gnosis staking subgraph indexes 58 contracts but only 41 are predict. The old `finalRoi` used the **unfiltered** `global.totalRewards` (over-counting 17 non-predict programs); staking rewards are now **filtered to the predict (Pearl) staking programs** (`PREDICT_STAKING_PROGRAMS_PEARL`, 12 entries) — the broader contract lists also include LST and other non-predict programs. Polygon's staking subgraph is predict-only, so it's summed unfiltered. **The displayed all-time ROI will drop accordingly on rollout** — this is a correction.

### Migration
`successRate`/`partialRoi`/`finalRoi` change shape (scalar → windowed). The predict snapshot category is bumped to `predict-v2` (rather than the global `METRICS_PREFIX`) to isolate the breaking blob change so the Brier, roi-distribution, and accuracy accumulators aren't reset.

## Validation
- `tsc --noEmit` and `eslint` pass (only pre-existing `no-console` warnings).
- Live-validated against subgraphs: Omen accuracy query (returns settled bets), Gnosis filtered staking rewards (~160k OLAS/30d over 22 cursor pages), Polygon unfiltered (~33k OLAS/90d).

## To verify on this Vercel preview
- Hit `/api/refresh-metrics/predict` and confirm `metrics.<platform>.{partialRoi,finalRoi,successRate}` are windowed objects. The **Polymarket accuracy query is unvalidated locally** (no subgraph URL in `.env`) — confirm it returns data here.
- First runs log `:backfilling` fetch-errors and only populate recent windows until the accumulators catch up (~hours, like Brier). The empty `predict-v2` blob means the predict page shows `--` for ~1 cron cycle, then fills in.

## Review follow-ups

- [ ] **Confirm the Polymarket accuracy query returns data on the preview** before merging (cannot be hit from local `.env`).
- [ ] After rollout, **delete the orphaned `metrics-{production,development}-predict.json` blobs** from the Vercel Blob dashboard (superseded by `predict-v2`).
- [ ] Follow-up: **APR** (`predict/index.ts` + `DataPage/OmenstratAprInfo`) still uses the broad staking-contract list — scope it to `PREDICT_STAKING_PROGRAMS_PEARL` too.
- [ ] Set **`NEXT_PUBLIC_GNOSIS_BALANCER_URL`** on the preview/prod env so omenstrat `finalRoi` (OLAS price) resolves.

_Deliberate (no change): the Performance card defaults to `Max` while the ROI distribution chart defaults to `7D` — headline shows the all-time figure, the chart is exploratory._


common-util/api/predict/staking-rewards.ts
```diff
@@ -0,0 +1,242 @@
...
+export type { WindowKey };
+
+// Omenstrat: filter to the 41 predict staking contracts (the gnosis staking subgraph
+// also indexes non-predict programs).
+export const fetchOmenstratStakingRewards = (): Promise<StakingRewardsWindows> =>
+  buildWindowedStakingRewards(
+    'predict-staking-rewards/omenstrat',
+    'gnosis',
+    OMEN_GENESIS_DAY,
+    GNOSIS_STAKING_CONTRACTS
```
Tanya-atatakai: ❗ **Change requested: this contract list includes non-predict (LST) programs.**

Filtering to `GNOSIS_STAKING_CONTRACTS` is a big improvement over the unfiltered `global.totalRewards`, but this list isn't fully predict either — some of these 41 staking contracts now belong to LST, not predict. That re-introduces the same over-counting this PR fixes, one level deeper: LST staking rewards would inflate predict final ROI.

Please narrow this to only the predict staking programs configured in Pearl. Note `PREDICT_STAKING_PROGRAMS_PEARL` already exists in `common-util/constants.ts:275` (12 entries, currently unused) — worth cross-checking it against Pearl's current config and using that as the source of truth here.

The APR calculation (`predict/index.ts:191` and `DataPage/OmenstratAprInfo`) uses the same 41-entry list, so it likely needs the same fix — fine as a follow-up, but the rewards path should be correct from the start since `max`-window numbers backfill into the accumulator.

components/AgentEconomies/PredictPage/PlatformActivitySection.tsx
```diff
@@ -143,15 +146,18 @@ export const PlatformActivitySection = ({
 }: PlatformActivitySectionProps) => {
   const m = metrics[platform];
 
-  // Brier score is the only windowed metric today, and it exists for Omenstrat only.
-  const isWindowed = platform === 'omenstrat' && !isNil(m.brierScore);
+  // ROI, Accuracy (both platforms) and Brier (Omenstrat) are windowed, so tabs are
+  // enabled whenever windowed data exists.
+  const isWindowed = !isNil(m.successRate) || !isNil(m.finalRoi);
   const [activeWindow, setActiveWindow] = useState<WindowKey>('max');
```
Tanya-atatakai, line 166: 💬 The Performance card defaults to `'max'` while `RoiDistributionChart` defaults to `'7d'`, and the two tab states aren't synced (only the platform switcher is). Probably fine, but worth a deliberate decision — landing on the page, the two cards answer different time ranges side by side.

common-util/api/predict/accuracy.ts
```diff
@@ -0,0 +1,319 @@
...
+      if (!answer || answer === INVALID_ANSWER_HEX) continue;
+      const day = dayOf(ts);
+      const cur = perDay.get(day) || { won: 0, total: 0 };
+      cur.total += 1;
+      if (Number(answer) === Number(bet.outcomeIndex)) cur.won += 1;
+      perDay.set(day, cur);
+    }
+
+    if (rows.length < LIMIT) break;
+    cursor = minTs; // next page strictly older
```
Tanya-atatakai: 💬 Cursor-pagination edge (also `accuracy.ts:175` and `staking-rewards.ts:93`): when a page returns exactly `LIMIT` rows, the next cursor is `timestamp_lt: minTs`, so any remaining rows sharing that exact timestamp are skipped. With second-granularity timestamps this is a rare, undercount-only edge, but it's silent. If you want to keep it, a one-line comment acknowledging the tradeoff would help; otherwise re-fetch with `timestamp_gte: minTs` overlap and dedupe, or page by `id_gt` within a timestamp.

common-util/api/predict/accuracy.ts
```diff
@@ -0,0 +1,319 @@
...
+const BACKFILL_CHUNK_DAYS = 30;
+
+// UTC-midnight genesis days, mirroring OMEN_GENESIS_TS / POLYMARKET_GENESIS_TS in
+// roi-distribution.ts (and OMEN_GENESIS_DAY in omenstrat-brier.ts). Backfill walks
+// down to here, no further.
+const OMEN_GENESIS_DAY = 1763769600;
+const POLYMARKET_GENESIS_DAY = 1768867200;
+
+// JSON-safe per-day bucket: settled bets and how many were correct.
+type AccuracyBucket = { won: number; total: number };
```
Tanya-atatakai, line 37: 💬 Note: accuracy buckets bets by *placement* day (consistent with Brier), while the windowed ROI inherits *settlement*-day bucketing from roi-distribution. Both are reasonable individually, but the same "7D" tab covers slightly different bet populations across the card's metrics. Not asking for a change — just confirm it's intentional, and it may deserve a sentence in the `/data` methodology text.

common-util/api/predict/index.ts
```diff
@@ -214,15 +212,25 @@ const fetchPolystratOlasApr = async (): Promise<MetricWithStatus<string | null>>
 
 export type { WindowKey, WindowedMetric } from './omenstrat-brier';
 
+// Snapshot category for the predict metrics blob. The `-v2` suffix is a one-time
+// version bump: partialRoi/finalRoi/successRate changed shape (scalar -> windowed
+// per-range), a breaking schema change. Bumping the category (rather than the global
+// METRICS_PREFIX) isolates the migration to this blob so the Brier, roi-distribution
+// and accuracy accumulators don't get reset. Reader and writer must use this.
+export const PREDICT_SNAPSHOT_CATEGORY = 'predict-v2';
```
Tanya-atatakai, line 220: 💬 Be aware of the backfill + `mergeWithFallback` interaction: while the accumulators backfill, the `:backfilling` fetch-error marks the metric stale, so `mergeWithFallback` keeps serving the *first* run's partially-filled window object rather than progressively filling windows; values snap in only once backfill completes (~hours). Matches the PR description's expectations — flagging it so nobody chases "the 30d tab isn't filling in" during catch-up.

common-util/api/predict/roi-distribution.ts
```diff
@@ -761,6 +761,86 @@ const computeAgentBlueprintHistogram = (
...
+ * aggregates into one ratio instead of binning — and intentionally OMITS the
+ * `MIN_TRADES_FOR_ROI_DISPLAY` activity filter: dollar weighting already de-emphasises
+ * tiny agents, and the Performance headline is a whole-economy figure, not a per-agent
+ * distribution. Keep the formulas in sync with the histogram.
+ *
+ * `daysBack` 7|30|90 → sum `byDay` buckets in the window; `null` → sum `allTimeAgents`.
+ * Results are 1e18-scaled (USDC 1e6 × 1e12 for polystrat). partialRoi% =
+ * netGain / totalCosts × 100; finalRoi adds staking rewards (USD) to the numerator.
+ */
+export const computeWindowedNetGainAndCosts = (
```
Tanya-atatakai, line 778: Nit: the 90d window's oldest day equals `BYDAY_RETENTION_DAYS`' prune cutoff exactly (prune is strict `<`, so it works today). If either constant ever changes independently, the 90d window silently loses days. Consider a comment tying the two, or asserting `BYDAY_RETENTION_DAYS >= 90` near the window definition.

Tanya-atatakai (commented):
Overall this is a solid change — the windowed ROI genuinely reuses the roi-distribution snapshot and mirrors `computeAgentBlueprintHistogram`'s math (same window convention: last N full UTC days ending yesterday), and the accumulator pattern matches the existing Brier one. The `predict-v2` category bump is correctly scoped so the preview validation flow can't clobber the production blob and the other accumulators aren't reset.

One change requested (staking contract list, see inline), plus two points on files outside this diff:

**`/data` methodology pages are now stale** — the Performance card links to `/data#<platform>-predict-accuracy` and `#<platform>-predict-roi`, but `components/DataPage/OmenstratAccuracy.tsx` / `PolystratAccuracy.tsx` still describe the old methodology ("based on the latest 10,000 trades/bets"), and the ROI info pages still describe the all-time, unfiltered-staking calculation. Since `/data` is the data-verification/provenance page, it should describe the new windowed, placement-day-bucketed accuracy and the windowed ROI (incl. the predict-contracts staking filter) — either in this PR or an immediate follow-up.

**`components/AgentEconomies/PredictPage/Activity.tsx:79` (nit)** — `roiStatus` comes from `finalRoi.status` only, so the partial-ROI value in the popover shows final ROI's staleness. Partial ROI can be fresh while final ROI is stale (rewards backfilling). Cheap fix: pass `partialRoi.status` separately for the popover; fine to skip if the popover is considered secondary.

Two post-merge follow-ups to add to the description:
- Delete the orphaned `metrics-production-predict.json` (and `-development-`) blob from the Vercel Blob dashboard after rollout.
- The Polymarket accuracy query is flagged as unvalidated — please confirm it returns data on the preview **before** merging, not after.

common-util/graphql/queries.ts
```diff
@@ -282,6 +282,80 @@ export const getExplorerBetsQuery = ({ first, timestamp_lt }) => gql`
...
+        number
+      }
+    }
+  }
+`;
+
+// Staking reward checkpoints within a [timestamp_gte, timestamp_lt) window, cursor-paged.
+// Powers the windowed staking-rewards accumulator used by windowed final ROI.
+// `contractAddresses` scopes to the predict staking contracts (omenstrat → the 41
+// GNOSIS_STAKING_CONTRACTS; polystrat → omit, the polygon staking subgraph is
```
Tanya-atatakai: Nit (comment-only): this still says omenstrat scopes to "the 41 GNOSIS_STAKING_CONTRACTS", but the caller now passes the 12 `PREDICT_STAKING_PROGRAMS_PEARL` addresses — this re-documents exactly the over-broad list the LST fix moved away from. Worth a one-line update so a future reader doesn't "restore" the 41-contract list believing it's the intended scope.

Tanya-atatakai (dismissed):
Re-reviewed — all points from the previous round are addressed:

- ✅ Staking rewards now scoped to `PREDICT_STAKING_PROGRAMS_PEARL` (verified against the constant — 12 Pearl predict programs), with the rationale documented in `staking-rewards.ts` and the `/data` methodology + example query updated to match.
- ✅ Cursor-pagination boundary tradeoff acknowledged at all three pager sites.
- ✅ Placement-day vs settlement-day bucketing documented as intentional (code comment + `/data` + accuracy popover).
- ✅ 90D / `BYDAY_RETENTION_DAYS` invariant comment added.
- ✅ Backfill × `mergeWithFallback` behavior and the `Max`-vs-`7D` default both declared deliberate in the PR body.

Also re-verified the windowed/all-time ROI math consistency (incl. the polystrat 1e6→1e18 scaling) and that all failure paths degrade to stale-with-fallback. One comment-only nit inline.

**Approving with a merge precondition**: please complete the PR's own preview checklist before merging — in particular **confirm the Polymarket accuracy query returns data on the Vercel preview** (unverifiable locally) and set `NEXT_PUBLIC_GNOSIS_BALANCER_URL` so omenstrat finalRoi resolves. The APR-scoping follow-up is fine as a separate PR.

common-util/graphql/queries.ts
mohandast52: Stale comment: the actual caller (`fetchOmenstratStakingRewards`) passes `PREDICT_STAKING_PROGRAMS_PEARL` (12 entries), not the 41-entry `GNOSIS_STAKING_CONTRACTS` — `staking-rewards.ts` even explains why the broader list would inflate ROI. Worth fixing so the two comments don't contradict each other.

common-util/api/predict/windowed-roi.ts
```diff
@@ -0,0 +1,116 @@
...
+      // a price; otherwise it stays null and the UI shows `--` for that tab.
+      const rewardsWei = rewards.value?.[key];
+      if (rewardsWei != null && olasUsdPrice) {
+        const rewardsUsdWei = (BigInt(rewardsWei) * olasUsdPrice) / SCALE_1E18;
+        final[key] = Number(((netGain + rewardsUsdWei) * 10000n) / totalCosts) / 100;
+      }
+    }
+  }
+
+  const partialStatus = createStaleStatus({
```
mohandast52, line 97: Suggestion (non-blocking): `partialRoi` reads the roi-distribution snapshot but its status only reflects whether the blob *loaded*, not how old it is. If the daily `predict-roi-distribution` cron fails for days, windowed ROI keeps publishing aging data with a fresh-looking status. Consider checking `snap.timestamp` and pushing a fetchError (e.g. when older than ~48h) so staleness propagates to the UI the same way it does for the subgraph-backed metrics.

components/AgentEconomies/PredictPage/PlatformActivitySection.tsx
```diff
@@ -143,15 +149,18 @@ export const PlatformActivitySection = ({
 }: PlatformActivitySectionProps) => {
   const m = metrics[platform];
 
-  // Brier score is the only windowed metric today, and it exists for Omenstrat only.
-  const isWindowed = platform === 'omenstrat' && !isNil(m.brierScore);
+  // ROI, Accuracy (both platforms) and Brier (Omenstrat) are windowed, so tabs are
+  // enabled whenever windowed data exists.
+  const isWindowed = !isNil(m.successRate) || !isNil(m.finalRoi);
```
mohandast52: Nit: `m.successRate` / `m.finalRoi` are now window *objects*, so this is truthy even when every window is `null` (e.g. fresh `predict-v2` blob mid-backfill) — the tabs enable and just show `--`, which contradicts the comment above ("When no windowed data is available yet, the non-max tabs stay disabled"). If the disabled-tabs behaviour is intended, check for at least one non-null window, e.g. `Object.values(m.finalRoi ?? {}).some((v) => v != null)`; otherwise just update the comment.

mohandast52 (commented):
Reviewed the full diff, verified the cross-references against the repo (genesis constants, `BYDAY_RETENTION_DAYS`, Pearl program list, roi-distribution category names), and ran `tsc --noEmit` + `eslint` on the branch — both clean (only pre-existing `no-console` warnings).

The accumulator design is solid: head-refresh/backfill logic mirrors `omenstrat-brier.ts` correctly, the cursor-paging undercount tradeoff is documented, windows are only published once fully covered (so `mergeWithFallback` keeps prior values), and the 90D window sits exactly on the `BYDAY_RETENTION_DAYS = 90` boundary as documented. The Pearl-program scoping fix for Gnosis staking rewards is a genuine correction. No dangling imports from the deleted modules; `stakingGlobalsQuery`/`getClosedMarketsBetsQuery` are still used elsewhere, so keeping them is right.

**One real issue found — the OG image pipeline still reads the old blob:**

`common-util/og/load-snapshots.ts` maps the OG `predict` category to the blob category `'predict'`:

```ts
const SNAPSHOT_CATEGORY: Record<OgSnapshotCategory, string> = {
  ...
  predict: 'predict',
```

After this PR the hourly cron only writes `predict-v2`, so the `agent-economies/predict` OG card (`OG_PAGE_REGISTRY` → `snapshots: ['predict']` → `predict.omenstrat?.dailyActiveAgents?.value`) will read a frozen blob — and will show `—` once the orphaned `metrics-*-predict.json` blobs are deleted per the PR's own follow-up checklist. Since the file isn't in the diff it can't get an inline comment: please point it at `PREDICT_SNAPSHOT_CATEGORY` (it only consumes `dailyActiveAgents`, whose shape is unchanged, so it's a one-line fix).

Remaining inline comments are minor (a stale doc comment and two suggestion-level robustness notes). Happy to approve once the OG snapshot category is updated.

mohandast52 (approved):
Re-reviewed at `a6c6a46`. Commit `8c919f9` cleanly addresses all three points from my last round — verified each against the new head, and `tsc --noEmit` + `eslint` both pass (only the pre-existing `no-console` warning).

**1. OG blob category (the blocker) — fixed.** `og/load-snapshots.ts` now maps the predict OG card to `PREDICT_SNAPSHOT_CATEGORY` instead of the literal `'predict'`, so the `agent-economies/predict` card reads the `predict-v2` blob the cron actually writes — no longer freezes when the orphaned `predict` blob is deleted per the rollout follow-up. Confirmed no other snapshot-category readers of the old `'predict'` blob remain (the two surviving `'predict'` strings are unrelated UI tab identifiers).

**2. ROI snapshot staleness — fixed.** `windowed-roi.ts` now reads `snap.timestamp` and pushes a `roi-distribution:<platform>:stale` fetchError once the blob is older than `ROI_SNAPSHOT_MAX_AGE_MS` (48h), so a stalled daily roi-distribution cron surfaces in the UI instead of publishing aging data with a fresh status. Cleanly threaded through both `partialStatus` and `finalStatus`.

**3. Tab gating — fixed.** The new `hasWindowData()` helper gates `isWindowed` on at least one non-null window (and now covers `partialRoi`/`finalRoi`/`successRate`/`brierScore`), so a fresh `predict-v2` blob mid-backfill correctly keeps the non-max tabs disabled — matching the `getTimeRangeTabs` comment.

**4. Stale doc comment — fixed.** `getStakingRewardsByTimeRangeQuery` now documents the 12 `PREDICT_STAKING_PROGRAMS_PEARL`, consistent with the caller.

LGTM — approving. The PR's own follow-up checklist (delete orphaned `metrics-{production,development}-predict.json` blobs after rollout; confirm the Polymarket accuracy query returns data on the preview; set `NEXT_PUBLIC_GNOSIS_BALANCER_URL`) still stands as post-merge ops.

Tanya-atatakai (approved):

PR #524, opened 2026-06-16
feat(fees): show on-chain mech marketplace fees collected

atepem:
## What

Show the **real Mech Marketplace protocol fees collected** across the site, computed on-chain, plus related homepage Activity polish.

### Fees collected (main change)
- New shared `fetchMechMarketplaceFeesCollected()` (`common-util/api/mech-marketplace-fees.ts`) reads `collectedFees()` directly from each chain's `BalanceTracker` and sums to USD. Scope today: USD-pegged tokens only — **USDC** (Ethereum, Arbitrum, Celo, Optimism, Polygon) and **Gnosis xDAI** — so no price oracle is needed.
- Wired into both snapshot pipelines (no new cron/endpoint): the **main** snapshot (`feesCollected`) and the **agent-economies** snapshot (`mechFees.protocolFees`).
- **Homepage Activity**: the "fees collected" tile now shows the real value instead of the previous `turnover × 15%` estimate (`MARKETPLACE_FEE_RATE` removed); USD formatting guards against `$NaN` when a snapshot predates the field.
- **agent-economies/mech**: the previously hardcoded "Marketplace Fees Burned" ($0) tile is now **"Fees Collected"** showing the computed value. `olasBurned` stays `0` — it's a separate flow and unit (OLAS burned on Ethereum) and the counted tokens are non-OLAS (they route to the Treasury).

### Data verification page
- Rewrote the Protocol Fees provenance copy and added per-chain `collectedFees()` on-chain verification links. The mech-page "Fees Collected" tile links to `/data#protocol-fees`.

### Homepage Activity layout fixes
- Fixed flywheel arrow alignment after the Users card was widened (the lower "ON" / "Agents are active" arrows now line up with the top arrows).
- Fixed the "OLAS currently staked" info-icon spacing/centering and the popover text overflow; narrowed the mech fee-flow popover.

### Earlier label changes (unchanged)
- **OLAS staked** → **OLAS currently staked**; **DAAs** → **Daily Active Agents**.

## Notes / follow-ups
- `collectedFees()` reflects fees **not yet drained** (it resets after each on-chain `drain()`); a cumulative figure via `Drained` events is a TODO.
- Non-USD-native (ETH/POL/CELO) and OLAS trackers are intentionally excluded until USD pricing is added (TODO in `mech-marketplace-fees.ts`).
- The homepage flywheel "OLAS burned" card is still a hardcoded `$0` (accurate today; revisit when OLAS burning is tracked).

## How to verify
- Hit `/api/refresh-metrics/main` and `/api/refresh-metrics/agent-economies` and inspect `feesCollected` / `protocolFees`; cross-check against `collectedFees()` on the trackers (currently only Gnosis xDAI is non-zero, ~$9).

Screenshot 2026-06-16 at 18 23 21

image

### DIFF
--- a/common-util/api/agent-economies/mech-fees.ts
+++ b/common-util/api/agent-economies/mech-fees.ts
@@ -6,6 +6,7 @@ import {
 } from 'common-util/graphql/metric-utils';
 import { legacyMechFeesTotalsQuery, newMechFeesTotalsQuery } from 'common-util/graphql/queries';
 import { WithMeta } from 'common-util/graphql/types';
+import { fetchMechMarketplaceFeesCollected } from 'common-util/api/mech-marketplace-fees';
 
 type MechFeesResult = WithMeta<{
   global: {
@@ -31,10 +32,14 @@ export const fetchMechFeeMetrics = async () => {
       keyof typeof MECH_FEES_GRAPH_CLIENTS
     >;
 
-    const allResults = await Promise.allSettled([
-      ...chainKeys.map((chain) => MECH_FEES_GRAPH_CLIENTS[chain].request(newMechFeesTotalsQuery)),
-      legacyMechFeesGraphClient.request(legacyMechFeesTotalsQuery),
-      ...chainKeys.map((chain) => getChainBlockNumber(chain)),
+    const [allResults, feesCollected] = await Promise.all([
+      Promise.allSettled([
+        ...chainKeys.map((chain) => MECH_FEES_GRAPH_CLIENTS[chain].request(newMechFeesTotalsQuery)),
+        legacyMechFeesGraphClient.request(legacyMechFeesTotalsQuery),
+        ...chainKeys.map((chain) => getChainBlockNumber(chain)),
+      ]),
+      // Protocol fees collected on-chain (read directly from the BalanceTrackers).
+      fetchMechMarketplaceFeesCollected(),
     ]);
 
     const newFeeResults = allResults.slice(
@@ -91,12 +96,22 @@ export const fetchMechFeeMetrics = async () => {
     const status = createStaleStatus({ indexingErrors, fetchErrors, laggingSubgraphs });
     const createMetric = (value: number) => ({ value, status });
 
+    // Protocol fees collected on-chain, in USD (USDC + xDAI ~= 1 USD trackers — see
+    // fetchMechMarketplaceFeesCollected). Carries its own fetch status.
+    const collectedFeesUsd = feesCollected.value !== null ? Number(feesCollected.value) : 0;
+    const collectedFeesMetric = { value: collectedFeesUsd, status: feesCollected.status };
+
     return {
       totalFees: createMetric(Number(inUsd.toFixed(6))),
       claimedFees: createMetric(Number(outUsd.toFixed(6))),
       recievedFees: createMetric(Number(outUsd.toFixed(6))),
       unclaimedFees: createMetric(Number(unclaimed.toFixed(6))),
-      protocolFees: createMetric(0),
+      protocolFees: collectedFeesMetric,
+      // OLAS burned is a separate flow (OLAS-denominated fees → burned on Ethereum) and a
+      // different unit from the USD collected fees above. The trackers we currently count
+      // (USDC, xDAI) are non-OLAS and route to the Olas Treasury, not burned, so nothing is
+      // burned from them. TODO: compute the burned OLAS amount once OLAS mech trackers are
+      // added (see fetchMechMarketplaceFeesCollected).
       olasBurned: createMetric(0),
     };
   } catch (error) {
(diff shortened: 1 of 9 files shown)


kupermind (approved):
Re-reviewed after `feat(fees): add temporary Drained-event scan to fees collected` — this addresses the main point from my earlier pass (the value would otherwise sawtooth to 0 after each `drain()`). Verified the new path on-chain; **LGTM**.

**Correctness checks (all pass):**
- `DRAINED_TOPIC0` = `keccak256("Drained(address,uint256)")` — matches `cast keccak` exactly. ✓
- Event shape: `Drained(address indexed token, uint256 collectedFees)` → `token` is the topic, the amount is the single non-indexed data word, so `BigInt(log.data)` is the drained amount. ✓
- `FEE_LIVE_SINCE_SEC` = 2026-06-15 06:00 UTC sits ~30 min before the Ethereum proposal execution (06:30:35 UTC) and the L2 bridge messages land later still — so the scan window always starts before activation on every chain. And no `Drained` event can predate it anyway: pre-activation `fee = 0` ⇒ `marketplaceFee` rounds to 0 ⇒ `collectedFees` never increments ⇒ `drain()` reverts on zero. So the scan is complete, not just bounded. ✓
- `collectedFees() + Σ Drained` = (undrained) + (drained) = lifetime realised fees, no double-count (collectedFees resets to 0 on each drain). ✓
- Chunks are non-overlapping/no-gap (`[start, start+CHUNK-1]`, `start += CHUNK`); whole-range first, chunk only on range-limit errors, other errors rethrown → falls back to undrained value; the drained scan is best-effort and degrades to `collectedFees()` on RPC failure rather than failing the tile. ✓

**Live cross-check (today ~11:00 UTC):** no `Drained` events exist on any counted tracker yet (nobody's drained the small post-activation amounts), so cumulative currently == `collectedFees()`. Tile resolves to **~$18.45** — Gnosis 18.4515 xDAI (up from ~$9 yesterday, so accrual is working) + $0 on the other counted chains. Polygon still $0, correctly (no mech has claimed via `processPayment` there since the fee went live; the USDC in that tracker is requester/mech balances, not fees).

**Non-blocking (already self-documented as TEMPORARY):** the log-scan window grows over time and re-scans each cron run — the file's TODO to replace it with a subgraph cumulative field is the right follow-up. Token scope (native POL/ETH/CELO + OLAS excluded) remains a documented TODO. Neither needs to hold up this PR.

CI green (lint, dependency audit, license gate, Vercel deploy). Approving.

Tanya-atatakai (approved):

mohandast52 (approved):
Reviewed the full diff. LGTM — approving. ✅

The core change is well-built and follows the repo's established `MetricWithStatus` / `mergeWithFallback` conventions: per-tracker reads via `Promise.allSettled`, graceful `null` bubbling when **all** trackers fail so the previous snapshot is preserved, and a best-effort Drained scan that falls back to the un-drained amount on any error. Wiring into both the `main` and `agent-economies` snapshots (no new cron/endpoint) is the right call.

A few things I checked:
- `DRAINED_TOPIC0` is correct — `keccak256("Drained(address,uint256)")` = `0xb2559daa…786c3`, matches exactly.
- `blockTimeSec = 43200 / lagLimit` is self-consistent with the documented per-chain block times in `CHAIN_CONFIG` (e.g. Ethereum 3600 → 12s, Gnosis 9000 → ~5s).
- `FEE_LIVE_SINCE_SEC = 1781503200` does decode to 2026-06-15 06:00 UTC. 👍
- `formatUsd` guard against `$NaN` for pre-field snapshots, and the `collected + drained` raw-units sum before a single `formatUnits` — both correct.

Non-blocking follow-ups (none need to hold the merge):

1. **`Drained` data-word decoding assumes `token` is indexed and `amount` is the only non-indexed word** (`BigInt(log.data)`). If `amount` were indexed, the scan fails safe (`BigInt('0x')` throws → falls back to un-drained); but if *both* were non-indexed, `log.data` is 64 bytes and `BigInt(log.data)` yields a massively inflated value. The topic hash is verified, but the indexing layout is currently untested in prod (no drains have happened yet — `collectedFees` is ~$9 on Gnosis and the Drained scan returns `0n` everywhere today). Worth a one-line confirmation against the BalanceTracker ABI before drains start flowing.

2. **The Drained scan window grows unbounded and re-runs every cron.** On fast-block chains (Arbitrum ~0.27s) the chunked path becomes many sequential `eth_getLogs` calls over time and could eventually pressure the 300s budget. Already labelled TEMPORARY with a subgraph-cumulative-field replacement planned — just flagging to prioritize that follow-up rather than let it ride.

3. **Minor:** `olasBurnedBranch` / `CheckOlasBurnt` in `FeeMetrics.tsx` now carry "fees collected" semantics — fine for the transitional state, but a rename would help future readers.

4. **Minor:** both snapshots perform the same on-chain reads + log scans independently (duplicate work across two crons). Acceptable given the snapshot model.

PR #527, opened 2026-06-22
feat(explorer): [BDMARK-2202] add Babydegen tab with Optimus & Modius agents

atepem:
## What

Adds the **Babydegen** economy to the Agent Economy Explorer, alongside the existing Predict tab. Babydegen splits into its two agents — **Optimus** (Optimism) and **Modius** (Mode) — via a sub-toggle, each with its own colour-coded heatmap.

| | Optimus | Modius |
|---|---|---|
| Chain | Optimism | Mode |
| Colour ramp | Optimism red `#FF0420` | Mode lime `#C9ED29` |
| Status | Ongoing | Phased out (2025-09-18) |

Each agent exposes three daily-series metrics: **Daily Active Agents**, **Total Transactions**, and **Assets Under Management (AUM)**.

## Data sources

- **DAA + Transactions** — registry subgraph `dailyAgentPerformances` (agentId 40) on Optimism + Mode (`activeMultisigCount`, `txCount`). Reuses the existing `explorerOmenstratSeriesQuery` (already parameterised by `agentIds`) via a new shared `pageRegistryRows()` helper.
- **AUM** — `dailyPopulationMetrics.totalFundedAUM` (plain USD) from the babydegen population-metrics subgraphs, per chain.
- Optimus and Modius are stored as **two separate `MetricWithStatus` snapshot entries** (`babydegenOptimus`, `babydegenModius`) so each self-heals independently — a flaky Mode run never blanks Optimus, and vice versa.

## Modius phase-out

Modius was retired on `MODIUS_FIXED_END_DATE_UTC` (2025-09-18). On-chain, agentId 40 on Mode shows a wind-down tail plus a post-retirement flat-value artifact (an identical ~20 active / 60 tx every day from late Feb 2026 onward — not organic). To avoid surfacing that:

- The Modius series (DAA/Transactions, matching how AUM already behaved) is **capped at the phase-out date**.
- That final cell renders as a **ringed circle** (instead of a rounded square) and, on hover, reads **"Modius phased out"**.

## Component changes

- `components/ExplorerPage/index.tsx` — made the Explorer **economy- and agent-aware**: per-economy metric config, an agent sub-toggle (shown only for multi-agent economies), and ramp-driven heatmap/legend colouring.
- `components/ExplorerPage/DaaCalendarHeatmap.tsx` — added a `levelColors` prop (red/lime ramps via `HEATMAP_RAMPS`), a `usd` tooltip kind, and `markerDate`/`markerLabel` props for the phase-out marker.
- `components/ExplorerPage/EconomySelector.tsx` — enabled the Babydegen tab.
- `common-util/api/explorer.ts` — `fetchBabydegenAgentSeries(chain)` + the two new snapshot entries.
- `pages/agent-economies/explorer.tsx`, `pages/api/refresh-metrics/explorer.ts` — pass per-agent series, report per-agent counts.

## Schema / rollout note

The `explorer` snapshot gains two **additive** keys (`babydegenOptimus`, `babydegenModius`) — backward compatible (the page falls back to empty series), so no `METRICS_PREFIX` bump. **The Babydegen tab will render empty until the `refresh-metrics/explorer` cron runs once after deploy** (or it's triggered manually). Production env must have live URLs for the Optimism/Mode registry + babydegen subgraphs, or the cron writes nulls.

## Testing

- `yarn lint` and `tsc --noEmit` pass.
- Verified all four subgraph endpoints return data; confirmed `totalFundedAUM` is plain USD (Optimus ≈ $12.8k, Modius ≈ $4.8k).


ROMAN-VALORY (dismissed):

Tanya-atatakai (dismissed):
Solid, ship-ready. Traced the data path end-to-end (subgraph clients → fetch → snapshot → `getStaticProps` → component); `tsc --noEmit` and ESLint both clean. No correctness bugs found.

**What's right:**
- Two separate snapshot entries (`babydegenOptimus`/`babydegenModius`) so `mergeWithFallback` self-heals each agent independently — a flaky Mode run can't blank Optimus.
- `pageRegistryRows()` extraction cleanly dedupes the Omenstrat paging loop and is correctly chain-agnostic.
- AUM is best-effort (`fetchBabydegenChainAum` → `[]` on failure) while the registry drives staleness; `Promise.all` only rejects on a registry failure, so an AUM hiccup degrades one tile instead of nulling the agent.
- Both `optimism` and `mode` are present in every map this touches (`REGISTRY_GRAPH_CLIENTS`, `BABYDEGEN_GRAPH_CLIENTS`, `CHAIN_CONFIG`, `getChainBlockNumber`) — no silent `undefined` client.
- Additive snapshot keys with empty-series fallback — backward compatible, no `METRICS_PREFIX` bump needed.

**Minor / non-blocking (inline comments below):**
1. PR description says the Modius series is "capped at the phase-out date", but commit 41980e13 extends it to end-of-2025 (`MODIUS_SERIES_END_TIMESTAMP`) with the marker held at 2025-09-18. Worth updating the description.
2. The marker comment in `DaaCalendarHeatmap.tsx` calls it the "series' end-point" — no longer true now that the series runs past it.
3. The "Modius phased out" tooltip is unreachable if the marked day has no data for the selected metric (hover handler is gated on `cell.hasData`). Fine in practice for DAA/tx; flag only if you want it guaranteed.
4. AUM uses a different window than DAA/transactions (`first:1000` desc, unpadded vs. 1500-day padded/trimmed). Both cover full history today so they line up — just informational.

Approving. None of these block merge.

components/ExplorerPage/DaaCalendarHeatmap.tsx
```diff
@@ -533,14 +578,17 @@ export const DaaCalendarHeatmap = ({
...
+            ? levelColors[levelFor(cell.count, localThresholds)]
             : cell.color;
         const waveDelay =
           WAVE_START_DELAY + Math.max(0, cell.col - (lastRealCol - VISIBLE_COLS)) * WAVE_STAGGER;
         const anim =
           reduced || !firstWave
             ? undefined
             : `heatmap-cell-in ${WAVE_DURATION}ms ${EASE_OUT} ${waveDelay}ms backwards`;
+        // Marker cell (e.g. an agent's phase-out day): drawn as a ringed circle so it
+        // stands out as the series' end-point, with its own tooltip copy.
```
Tanya-atatakai: Nit: this comment calls the marker the "series' end-point," but since commit 41980e1 the Modius series extends to end-of-2025 while the marker stays at 2025-09-18 — so it's now a mid-series ring, not the end-point. Worth tweaking the wording.

components/ExplorerPage/DaaCalendarHeatmap.tsx
```diff
@@ -533,14 +578,17 @@ export const DaaCalendarHeatmap = ({
...
+        const isMarker = markerDate != null && cell.inRange && cell.date === markerDate;
```
Tanya-atatakai, line 600: Minor edge case re: the marker tooltip: the "<agent> phased out" label is only reachable if the marked day has data. Hover is gated on `cell.hasData` (the `onMouseEnter` handler below) and `onCellEnter` early-returns on `!cell.hasData`. The ring always draws, but if the marker day has no value for the selected metric (e.g. AUM) the label can't be hovered. Fine in practice for DAA/transactions — flag only if you want the label guaranteed.

components/ExplorerPage/DaaCalendarHeatmap.tsx
atepem: Good catch — reworded in 6e28566: the comment now describes it as "a notable day" that can sit mid-series, and notes the phase-out date is mid-series once a wind-down tail follows.

components/ExplorerPage/DaaCalendarHeatmap.tsx
atepem: Hardened in 6e28566. Marker cells are now hoverable even without a value for the active metric (onCellEnter allows them through, and the cell stays hoverable), and the tooltip omits the value line when the cell has no data — so on the AUM tile the "… phased out" label is always reachable even if that day has no AUM snapshot.

Tanya-atatakai (approved):

PR #543, opened 2026-07-13
fix(predict): stop roi-distribution runs from silently persisting failed fetches

atepem:
## Why

After the `METRICS_PREFIX` bump in #542 reset all snapshot blobs, Total ROI and the ROI-distribution histograms on `/agent-economies/predict` went blank for ~3 days — with a fresh-looking, non-stale status.

Root cause (verified against the prod blobs): the roi-distribution rebuild runs swallow every subgraph error (`catch` → `break` → empty result). One "successful" run advanced `lastDayTimestamp` 30 days while writing empty `byDay` data, and each run overwrote `allTimeAgents` with `{}` when the traderAgents fetch failed mid-rebuild. Empty data is indistinguishable from healthy data to the staleness system, so the UI confidently rendered `--`.

(Prod data has already been restored out-of-band by copying the healthy pre-bump blobs to the new prefix names and re-triggering the refresh endpoints — ROI is live again. This PR prevents the failure mode from recurring.)

## What

- **Daily-stats fetchers report `ok`**: a failed chunk no longer advances the day cursor, so it's retried on the next run instead of becoming a permanent hole in `byDay`.
- **`fetchAllTimeAgents` reports `ok`**: on failure the previous run's totals are kept instead of being replaced by a truncated/empty map (which nulls the Max-window ROI and the all-time histogram).
- **Failures are persisted** on the blob as a new optional `fetchErrors` field and surfaced by `windowed-roi.ts` as `MetricWithStatus` fetchErrors — plus a new `:backfilling` flag when the `byDay` cursor is ≥2 days behind. The UI stale indicator now fires, and `mergeWithFallback` preserves the last valid ROI instead of dropping to `--`.
- **Refresh endpoint echoes `fetchErrors`** in its JSON response for cron observability.

`fetchErrors` is an optional addition to the blob schema — existing blobs parse unchanged, so no `METRICS_PREFIX` bump is needed.

## Verification

- `yarn lint`, `tsc --noEmit`, `yarn build` all pass.
- Prod recovery path exercised the unchanged happy path end-to-end: restored blobs + manual endpoint triggers repopulated `allTimeAgents` (2,085 omenstrat / 174 polystrat) and windowed ROI on the live page.

### DIFF
--- a/common-util/api/predict/roi-distribution.ts
+++ b/common-util/api/predict/roi-distribution.ts
@@ -488,9 +488,9 @@ const updateAgentBlueprintData = async (
 
     if (!statsOk) {
       // Incomplete stats: apply nothing and do NOT advance lastDayTimestamp. The
-      // cursor only ever moves forward, so advancing past a failed fetch would
-      // leave a permanent hole in byDay (this is what emptied the blob after the
-      // 2026-07-08 prefix reset). The whole chunk is retried on the next run.
+      // day cursor only ever moves forward, so advancing past an incomplete fetch
+      // would permanently exclude those days from byDay. Leaving the cursor in
+      // place means the whole chunk is retried on the next run.
       runFetchErrors.push('daily-stats');
     } else {
       const statsByDay = new Map<string, DailyStatEntry[]>();
--- a/common-util/api/predict/roi-distribution.ts
+++ b/common-util/api/predict/roi-distribution.ts
@@ -93,6 +93,14 @@ export type AgentBlueprintRoiData = {
   >;
   lastDayTimestamp: number;
   allTimeAgents: Record<string, AllTimeAgentEntry>; // agentId → all-time totals
+  /**
+   * Subgraph fetches that failed during the run that wrote this blob
+   * ('daily-stats' | 'all-time-agents' | 'mech-requests'). Consumers (windowed
+   * ROI) surface these as MetricWithStatus fetchErrors so the UI can flag
+   * staleness — a blob written by a failing run is otherwise indistinguishable
+   * from a healthy one.
+   */
+  fetchErrors?: string[];
 };
 
 // ─── Internal query types ────────────────────────────────────────────────────
@@ -138,10 +146,15 @@ type SenderEntry = {
 
 // ─── Daily stat fetchers ─────────────────────────────────────────────────────
 
+// `ok: false` means a page failed mid-pagination — the stats are incomplete and
+// the caller must NOT advance its day cursor past this range, or the failed days
+// would be skipped forever (the cursor only moves forward).
+type DailyStatsResult = { stats: DailyStatEntry[]; ok: boolean };
+
 const fetchOmenstratDailyStats = async (
   dayStartTs: number,
   dayEndTs: number
-): Promise<DailyStatEntry[]> => {
+): Promise<DailyStatsResult> => {
   const results: DailyStatEntry[] = [];
   let skip = 0;
   while (true) {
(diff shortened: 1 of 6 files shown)


common-util/api/predict/roi-distribution.ts
```diff
@@ -451,121 +484,129 @@ const updateAgentBlueprintData = async (
...
-      if (!byDay[dKey]) byDay[dKey] = { agents: {} };
-      if (!byDay[dKey].agents[aid]) {
-        byDay[dKey].agents[aid] = { profit: '0', payout: '0', mechRequests: 0 };
+    const { stats: allStats, ok: statsOk } = await fetchStats(startDay, processEndDay);
+
+    if (!statsOk) {
+      // Incomplete stats: apply nothing and do NOT advance lastDayTimestamp. The
+      // cursor only ever moves forward, so advancing past a failed fetch would
+      // leave a permanent hole in byDay (this is what emptied the blob after the
+      // 2026-07-08 prefix reset). The whole chunk is retried on the next run.
```
Tanya-atatakai: Please drop the historical framing here. This comment explains the guard by narrating a past incident (a dated prefix reset / what previously emptied the blob). Rewrite it to describe only the invariant the code enforces now and why the line must stay — i.e. the day cursor only moves forward, so it must not advance past an incomplete fetch. No incident names, dates, or example triggers.

common-util/api/predict/windowed-roi.ts
```diff
@@ -64,11 +66,24 @@ const computePlatformWindowedRoi = async (
...
+    roiFetchErrors.push(`roi-distribution:${source}`);
+  } else {
+    if (roiSnapshotStale) roiFetchErrors.push(`roi-distribution:${source}:stale`);
+    // Subgraph failures recorded by the daily refresh run that wrote the blob
+    // (e.g. the all-time agents fetch failed and the previous totals were kept).
+    for (const err of roiData.fetchErrors ?? []) {
+      roiFetchErrors.push(`roi-distribution:${source}:${err}`);
+    }
+    // A byDay cursor ≥2 days behind means the blob is mid-backfill (e.g. after a
+    // METRICS_PREFIX reset) or the daily cron keeps failing — either way the
```
Tanya-atatakai: Same here — drop the historical framing (the `METRICS_PREFIX reset` aside). Keep only the current rule and why it matters: a blob whose day cursor is behind is computed from incomplete data and should be flagged. No reference to when or why it broke before, and no example triggers.

Tanya-atatakai (commented):
A few review notes. Two are on files outside this PR's diff, so they're here in the body rather than inline.

---

### Scope `METRICS_PREFIX` per category, not globally
**`common-util/snapshot-storage.ts:11-14`**

The version suffix is applied to every blob filename through one shared `METRICS_PREFIX`:

```js
const METRICS_PREFIX = `metrics-${process.env.NODE_ENV}-20260708`;
const getSnapshotFilename = (category) => `${METRICS_PREFIX}-${category}.json`;
```

This couples all categories to one version. A breaking schema change in one category (e.g. adding a metric to `other`) forces a bump that renames **every** blob — including all `predict-roi-distribution-*` snapshots — so unrelated categories return `null` and restart from genesis. Re-backfilling the entire predict ROI history because `other` changed shape is wrong.

Suggest versioning per category instead of one global prefix — e.g. a per-category version map folded into `getSnapshotFilename`, so a bump only invalidates the category that actually changed. A default/base version keeps categories that never change stable.

---

### Default the performance time-range tab to 7D
**`components/AgentEconomies/PredictPage/PlatformActivitySection.tsx:166`**

The performance section opens on **Max**:

```js
const [activeWindow, setActiveWindow] = useState<WindowKey>('max');
```

Please make **7D** the default selected tab in this PR — change the initial state to `'7d'`.

common-util/api/predict/roi-distribution.ts
atepem: Reworded in 98cdc9c3 — the comment now states only the invariant (the day cursor only moves forward, so it must not advance past an incomplete fetch; the chunk is retried next run).

common-util/api/predict/windowed-roi.ts
atepem: Reworded in 98cdc9c3 — now describes only the current rule: a cursor ≥2 days behind means the windowed values come from incomplete data, with 1 day of tolerance for the midnight→cron gap.

Tanya-atatakai (approved):

ROMAN-VALORY (approved):

PR #548, opened 2026-07-27
feat: read per-request mech data from mech-analytics behind a flag

Tanya-atatakai:
Resolves BDMARK-2238

## Summary
 
Mech requests move off-chain. After the switch, the marketplace subgraph does not create individual `Request` / `ParsedRequest` entities anymore — per-request reads (question title, tool) return nothing. Only two read paths on the website use per-request data:

1. The incremental mech-request feed for Partial ROI (`roi-distribution.ts`)
2. The per-sender tool pulls for tool accuracy (`tool-accuracy.ts`)

This PR moves both to the mech-analytics `/v1/data/scored-rows` endpoint, behind one flag. `USE_MECH_ANALYTICS` defaults to off, so this merge changes nothing in production.

Full design: `docs/mech-analytics-migration.md`.

## Changes

- **New** `common-util/api/predict/mech-analytics.ts` — endpoint client, the flag, the QMR fetch (rebuild + incremental), and a per-requester fetch. Incremental runs watermark on `computed_at` and remove duplicates with a persisted request-id ledger
- **`roi-distribution.ts`** — flag branch in `updateAgentBlueprintData`; `QmrData` gets optional `lastComputedAt` + `ingestedRequestIds` (additive, no schema-version bump); first flag-on run rebuilds the open set
- **`tool-accuracy.ts`** — flag branch; the duplicated gnosis/polygon fetchers are now one shared function with a `chain` parameter
- **`queries.ts`** — removed the now-unused `getMechRequestsBySenderEntityQuery`
- **`docs/mech-analytics-migration.md`** — spec with status header; indexed from `CLAUDE.md`
- **`.env.example`** — `MECH_ANALYTICS_URL`, `USE_MECH_ANALYTICS=false`

## What does NOT change

- `senderTotal` stays on the marketplace subgraph — `totalLegacyRequests + totalMarketplaceRequests` keeps growing at settlement time
- All aggregate metrics (homepage ATA/fees, mech economy tiles, explorer series, categorized request counts) stay on their current sources
- Downstream ROI logic: `fetchAllTimeAgents`, `openQmr` matching, `normalizeTitle`, `settledMechRequests`, byDay, TTL flush
- The tool-accuracy metric definition ("trader bets grouped by tool used, fraction won")

## Rollout

1. Merge with the flag off — no behavior change
2. Turn the flag on in a Vercel preview; run the checks from the spec's Verification section (flag-on numbers must equal flag-off numbers for the same window)
3. Turn the flag on in production at cut-over; watch 24–48h; roll back = flag off
4. Cleanup PR later: remove the flag and the subgraph code paths

## Test plan

- [x] `yarn lint` clean
- [x] `yarn build` passes
- [ ] Flag-off: refresh endpoints produce the same snapshots as main (code path unchanged)
- [ ] Flag-on (preview): `/api/refresh-metrics/predict-roi-distribution?agent=omenstrat|polystrat` twice — first run logs `rebuild ok=true`, second run logs `incremental` and ingests only new rows
- [ ] Flag-on (preview): `/api/refresh-metrics/predict-tool-accuracy` — per-tool counts match the flag-off run
- [ ] Flag on without `MECH_ANALYTICS_URL`: `'mech-requests'` fetchError, watermark does not advance

## References

- Migration plan: valory-xyz/mech-analytics `docs/consumer_migration.md`
- `since_computed_at` cursor: mech-analytics `etl/api/routes/data.py`
- Subgraph settlement handler (aggregate counters keep working): `autonolas-subgraph/subgraphs/marketplace/src/marketplace/mech-marketplace.ts` (`handleMarketplaceDeliveryWithSignatures`)


common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  console.log(
+    `[mech-analytics:${chain}] ${isRebuild ? 'rebuild' : 'incremental'} ok=${ok} ` +
+      `ingestedIds=${Object.keys(ingestedRequestIds).length}`
+  );
+
+  return {
+    additions,
+    rebuild: isRebuild,
+    lastComputedAt: isRebuild ? runStart : maxComputedAt,
```
DIvyaNautiyal07: **[Blocking]** On an incremental failure, this returns `lastComputedAt: maxComputedAt` — the mid-flight watermark accumulated up to the last successful page. Back in `roi-distribution.ts:479`, `newLastComputedAt = result.lastComputedAt ?? existingQmr?.lastComputedAt` treats that non-null value as authoritative and persists it. The next run uses it as `since_computed_at`, so every row from the failed pages is permanently skipped. `ok=false` only surfaces via StaleIndicator — the data loss is silent.

**Fix:** on incremental failure, return `lastComputedAt: null` (same as the rebuild-fail path) so `??` falls back to the prior watermark. The `ingestedRequestIds` map dedupes any overlap on the retry.

common-util/api/predict/tool-accuracy.ts
```diff
@@ -81,15 +82,39 @@ async function fetchResolvedBets(): Promise<Bet[]> {
...
+  chain: 'gnosis' | 'polygon',
   sender: string,
   timestampGt: number
 ): Promise<MechRequest[]> {
   const allRequests: MechRequest[] = [];
 
   for (let i = 0; i < MECH_MAX_PAGES; i++) {
     const skip = i * MECH_PAGE_SIZE;
-    const data = await MARKETPLACE_GRAPH_CLIENTS.gnosis.request<{
+    const data = await MARKETPLACE_GRAPH_CLIENTS[chain].request<{
```
DIvyaNautiyal07, line 124: **[P1]** The old `fetchPolygonMechRequestsForSender` used `getMechRequestsBySenderEntityQuery` (`sender(id: ...) { requests }`) and read `data.sender?.requests`. This new shared fetcher uses `getMechRequestsBySenderWithToolQuery` (top-level `requests(where: { sender: ... })`) and reads `data.requests` for **both** chains — including Polygon on the flag-off path (line 127 `if (!data.requests`).

If the Polygon marketplace subgraph doesn't support the top-level `requests` query with the sender-filter shape, `data.requests` will be `undefined`, the guard exits on page 1, and polystrat tool-accuracy silently returns empty **in production today** (flag off), not just after the flip.

**Fix:** verify the Polygon marketplace subgraph schema supports this query. If it doesn't, keep the entity-shape fetcher for Polygon.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,211 @@
...
+      url.searchParams.set(key, value);
+    }
+    if (cursor) url.searchParams.set('cursor', cursor);
+
+    const response = await fetch(url.toString());
+    if (!response.ok) {
+      throw new Error(`mech-analytics /v1/data/scored-rows returned ${response.status}`);
+    }
+    const page = (await response.json()) as ScoredRowsPage;
+    yield page.rows ?? [];
```
DIvyaNautiyal07: **[P1]** Docstring on `iterateScoredRows` says "throws if any request fails." But `page.rows ?? []` yields an empty array on a 200 response missing the `rows` key (proxy error / schema drift). Downstream this reads as "clean empty page" rather than "endpoint misbehaving" — the loop then breaks on the absent `next_cursor` and the whole call silently succeeds with zero rows.

**Fix:** raise if `payload.rows` isn't an array. Match the pattern the mech-analytics client added earlier this migration (`isinstance(rows, list)` check with a typed error).

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,211 @@
...
+    ok,
+  };
+};
+
+/**
+ * Loads all scored rows for one requester made after the given time
+ * (unix seconds). Keeps no state between runs.
+ * `since` filters on requested_at; +1 second turns ">=" into ">".
+ */
+export const fetchScoredRowsForRequester = async (
```
DIvyaNautiyal07, line 247: **[P1]** `fetchScoredRowsForRequester` truncates at `maxPages` without a log. Silent, but there's also a subtler ordering issue: the endpoint returns rows `ORDER BY computed_at, request_id ASC` (verified in mech-analytics HEAD `etl/api/routes/data.py:202`). The old subgraph paginated by `blockTimestamp` (request-time order).

At `MECH_MAX_PAGES=10 * MECH_PAGE_SIZE=1000 = 10k rows`, a sender with more than 10k requests gets earliest-by-scoring-order rows, not earliest-by-request-time. For senders that hit a backfill surge, the request-order and scoring-order sets diverge, and tool-accuracy under-counts silently.

**Fix:** log a warning when the cap is hit; consider a higher cap for the tool-accuracy use case, or fall back to the subgraph for that sender.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,211 @@
...
+ */
+export const fetchMechRequestsFromAnalytics = async (
+  chain: 'gnosis' | 'polygon',
+  lastComputedAt: string | undefined,
+  previousIngestedIds: Record<string, number> | undefined,
+  previousLastTimestamp: number
+): Promise<AnalyticsQmrUpdate> => {
+  const nowSec = Math.floor(Date.now() / 1000);
+  const windowStartSec = nowSec - QMR_WINDOW_SECONDS;
+  const isRebuild = !lastComputedAt;
```
DIvyaNautiyal07: **[P1]** `isRebuild = !lastComputedAt` treats any truthy string as a valid incremental baseline. If a corrupted / old-schema value ever gets persisted (e.g. `"undefined"` as a string, or a stale value from a previous schema), the incremental branch fires with a bogus `since_computed_at`. Endpoint either 400s (silent loop of `ok=false` with watermark stuck) or interprets as very-old date (floods the ID map).

**Fix:** validate before use. Something like `isRebuild = !lastComputedAt || Number.isNaN(Date.parse(lastComputedAt))`, and warn when the second condition trips.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,211 @@
...
+    }
+    const page = (await response.json()) as ScoredRowsPage;
+    yield page.rows ?? [];
+
+    if (!page.next_cursor) break;
+    cursor = page.next_cursor;
+  }
+}
+
+export type AnalyticsQmrUpdate = {
```
DIvyaNautiyal07: **[P2]** `AnalyticsQmrUpdate` encodes the failure invariant in a comment (`lastComputedAt: null` means rebuild failed, save nothing) rather than in the type. A callsite can access every data field regardless of `ok`, and the compiler won't complain.

**Fix:** discriminated union — `type AnalyticsQmrUpdate = { status: 'ok'; ... } | { status: 'failed' }`. TypeScript then narrows at every callsite, and the "save nothing on failure" invariant becomes a type error if violated.

.env.example
```diff
@@ -82,3 +82,7 @@ CELO_RPC=__URL__
 POLYGON_RPC=__URL__
 MODE_RPC=__URL__
 SOLANA_RPC=__URL__
+
+# mech-analytics API (off-chain migration, server-only) — see docs/mech-analytics-migration.md
+MECH_ANALYTICS_URL=__URL__
```
DIvyaNautiyal07, line 87: **[P2]** `base.replace(/\/$/, '')` in `mech-analytics.ts:48` strips a trailing slash but doesn't strip a path component. If someone sets `MECH_ANALYTICS_URL=https://api.example.com/v1` (common for versioned API gateways), the constructed URL becomes `https://api.example.com/v1/v1/data/scored-rows` and 404s.

**Fix:** one-line note here — `# Root URL only, do not include /v1 or any path`.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,211 @@
...
+        if (!Number.isFinite(ts) || ts <= 0 || ts < windowStartSec) continue;
+
+        if (isRebuild && row.resolution_status === 'invalid') {
+          ingestedRequestIds[row.request_id] = ts; // remember the id so we skip this row later
+          continue;
+        }
+
+        const agentId = row.requester?.toLowerCase();
+        const questionTitle = row.question_title;
+        if (!agentId || !questionTitle) continue;
```
DIvyaNautiyal07: **[P2]** Three silent skip paths — line 131 (`ts <= 0` / out-of-window), line 140 (`!agentId || !questionTitle`), line 133 (`resolution_status === 'invalid'` during rebuild) — all `continue` with no counter and no log. A regression in the endpoint (e.g., a batch shipping `requester: null` everywhere) is invisible until `openRequestCount` drifts.

**Fix:** sampled log or a counter dict returned alongside `AnalyticsQmrUpdate`, surfacing in the same place as `fetchErrors`.

common-util/api/predict/roi-distribution.ts
```diff
@@ -55,6 +57,11 @@ const dayKeyOf = (ts: number): string => String(Math.floor(ts / DAY_SECONDS) * D
 export type QmrData = {
   questionMechRequests: Record<string, Record<string, number[]>>; // title → agentId → sorted asc timestamps
   lastMechRequestTimestamp: number;
+  // mech-analytics path only — see docs/mech-analytics-migration.md
+  /** Last seen computed_at. When missing, the next flag-on run rebuilds the open set. */
+  lastComputedAt?: string;
```
DIvyaNautiyal07: **[P2]** `lastComputedAt?: string` and `ingestedRequestIds?: Record<...>` are added as independent optionals — the type permits three states (neither, one, both) but the domain invariant is "both arrive together from a successful mech-analytics run, or neither." A consumer reading `lastComputedAt` and assuming `ingestedRequestIds` is present will crash at runtime.

**Fix:** nest into a single object — `ingestionMeta?: { lastComputedAt: string; ingestedRequestIds: Record<string, number> }` — so "both or neither" becomes structural.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,211 @@
...
+
+const PAGE_SIZE = 5000; // endpoint max limit
+
+export const MECH_ANALYTICS_CHAIN_IDS: Record<'gnosis' | 'polygon', number> = {
+  gnosis: 100,
+  polygon: 137,
+};
+
+/** Same value as QMR_MAX_AGE_DAYS in roi-distribution.ts. Older requests cannot be open anymore. */
+const QMR_WINDOW_SECONDS = 14 * 86400;
```
DIvyaNautiyal07: **[Nit]** `QMR_WINDOW_SECONDS` is bound to `QMR_MAX_AGE_DAYS` in `roi-distribution.ts` by comment reference, with no import or assertion. If either value changes, the other silently drifts. Consider exporting one from a single source (e.g., `QMR_MAX_AGE_DAYS` from `roi-distribution.ts` and computing `14 * 86400` here from that).

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,211 @@
...
+  searchParams: Record<string, string>
+): AsyncGenerator<ScoredRow[]> {
+  const base = process.env.MECH_ANALYTICS_URL;
+  if (!base) {
+    throw new Error('MECH_ANALYTICS_URL not set — cannot fetch mech-analytics scored rows');
+  }
+
+  let cursor: string | null = null;
+  while (true) {
+    const url = new URL(`${base.replace(/\/$/, '')}/v1/data/scored-rows`);
```
DIvyaNautiyal07: **[Nit]** URL construction + fixed-param loop run on every page. Hoist `base.replace(...)` and the `Object.entries(searchParams)` set-calls above the `while (true)`; only the `cursor` set-call needs to be inside. Trivial cost, readability win — the loop body becomes just cursor-per-page state.

common-util/api/predict/tool-accuracy.ts
Tanya-atatakai: the subgraphs are identical in that term, I verified of course 🙏

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Checked this against the endpoint SQL (`etl/api/routes/data.py`): rows are served `ORDER BY computed_at, request_id` ascending, and `since_computed_at` uses `>=` (inclusive). `maxComputedAt` only advances over rows we processed, and pages are atomic. So every row on a failed page has `computed_at >= maxComputedAt`. The next run starts at that watermark and receives the failed pages again; `ingestedRequestIds` removes the boundary duplicates. No row can be skipped, so no code change here. Documented the proof in docs/mech-analytics-migration.md (cfe4890d).

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in cfe4890d — `iterateScoredRows` now throws when a 200 response has no `rows` array.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in cfe4890d — the analytics page size for tool accuracy is now 5000 (cap: 50k rows per sender, 5x the subgraph cap), and `fetchScoredRowsForRequester` warns when the cap cuts rows. No subgraph fallback: that would keep the dependency the migration removes.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in cfe4890d — a watermark that does not parse triggers a warn and a rebuild.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in cfe4890d, with a different shape than proposed. A `{ status: 'failed' }` variant without data would drop the partial additions that an incremental failure must keep (same partial-keep behavior as the subgraph path). Instead the fetch returns `null` on rebuild failure only — 'save nothing' is now structural — and `lastComputedAt` no longer needs `null`.

.env.example
Tanya-atatakai: Declined. Repo convention is base URLs in env (`OLAS_API_URL`, `NEXT_PUBLIC_AFMDB_URL`, all subgraph URLs) and the code owns the paths — the client appends `/v1/data/scored-rows` itself. A wrong URL 404s and surfaces through the named fetchError, the normal failure path.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in cfe4890d — skips are counted per reason (dup / old / invalid / noKey) and appear in the per-run log line.

common-util/api/predict/roi-distribution.ts
Tanya-atatakai: Fixed in cfe4890d — the fields are now nested as `QmrData.mechAnalytics`, so both-or-neither is structural. The fields are not in any production blob yet, so the shape change is free.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in cfe4890d — `QMR_MAX_AGE_DAYS` moved to `common-util/constants.ts` and both files import it. Note: exporting it from `roi-distribution.ts` would create an import cycle (`roi-distribution` already imports `mech-analytics`, and the constant is used at module top level), so the shared-constants module is the safe home.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in cfe4890d — the URL and fixed params are built once; only `cursor` changes per page.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,244 @@
...
+    params.resolved = 'false';
+  } else {
+    params.since_computed_at = lastComputedAt as string;
+  }
+
+  try {
+    for await (const rows of iterateScoredRows(params)) {
+      for (const row of rows) {
+        if (row.computed_at && (!maxComputedAt || row.computed_at > maxComputedAt)) {
+          maxComputedAt = row.computed_at;
```
DIvyaNautiyal07, line 168: **[P1]** `row.computed_at` is used unvalidated: line 146 does a lexicographic string compare (`row.computed_at > maxComputedAt`), which is only safe if every value is a well-formed ISO 8601 string. A malformed value from the API (`"null"`, a naive date, garbage) becomes the new `maxComputedAt`, gets saved at line 205, and comes back next run as `since_computed_at`. If the API then 422s on it, `iterateScoredRows` throws → `ok=false` → but the incremental path still saves the bad watermark (line 205 uses `maxComputedAt as string`), so the cursor pins permanently.

The rebuild path self-heals (uses `runStart` at line 205), the incremental path doesn't. Suggest guarding with `Number.isFinite(Date.parse(row.computed_at))` alongside the truthiness check, or skip/warn on malformed values so a bad row can't poison the watermark.

common-util/api/predict/roi-distribution.ts
```diff
@@ -445,11 +457,46 @@ const updateAgentBlueprintData = async (
...
+      existingQmr?.mechAnalytics?.ingestedRequestIds,
+      newMechTs
+    );
+    // null = the rebuild failed; keep everything as it is and retry next run.
+    mechRequestsOk = result !== null && result.ok;
+    if (result !== null) {
+      if (result.rebuild) {
+        // Old entries from the subgraph cannot be matched with analytics rows,
+        // so we drop them and start fresh.
+        qmr = {};
```
DIvyaNautiyal07, line 482: **[P1]** `qmr = {}` fires whenever `result.rebuild` is true, regardless of row count. A successful-but-empty rebuild (misconfigured `chain_id`, endpoint returning `{"rows":[]}` while catching up, brief upstream data-load issue) wipes the whole open set. Since `settledMechRequests = senderTotal - openRequestCount` reads from `senderTotal` (which keeps growing on the subgraph), a wiped `openRequestCount` then silently over-inflates settled and undercounts open until real rows land.

Suggest refusing to wipe when `Object.keys(result.additions).length === 0` and logging loudly, or making "empty rebuild is real" require an explicit signal from the fetcher (e.g. a distinct `emptyOnPurpose` flag from the API's own "no data yet" case).

common-util/api/predict/tool-accuracy.ts
```diff
@@ -81,15 +84,39 @@ async function fetchResolvedBets(): Promise<Bet[]> {
...
       requests: MechRequest[];
     }>(
       getMechRequestsBySenderWithToolQuery({
```
DIvyaNautiyal07, line 127: **[P1]** The deleted `getMechRequestsBySenderEntityQuery` used `sender(id:…) { requests }` on polygon. The consolidated function now uses `getMechRequestsBySenderWithToolQuery` — `requests(where:{ sender:… })` — for both chains. If the polygon marketplace subgraph doesn't expose that same top-level `requests` entity with a `sender` field filter (or wires them differently), polygon flag-OFF silently returns empty rows and polystrat tool-accuracy zeroes out — this is a live path today, not gated by `USE_MECH_ANALYTICS`.

Could you confirm the polygon marketplace subgraph shape matches gnosis on this query before merge? A quick check against the polygon subgraph playground would settle it.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,244 @@
...
+
+/**
+ * Client for the mech-analytics `/v1/data/scored-rows` endpoint.
+ * It replaces the per-request reads from the marketplace subgraph,
+ * because the subgraph has no data for off-chain requests.
+ * Full details: docs/mech-analytics-migration.md
+ */
+
+/** One switch that turns on all mech-analytics read paths. */
+export const USE_MECH_ANALYTICS = process.env.USE_MECH_ANALYTICS === 'true';
```
DIvyaNautiyal07, line 11: **[P1]** `USE_MECH_ANALYTICS` gates two branches (this file's consumers in `roi-distribution.ts` L468 and `tool-accuracy.ts` L143). There's no automated test that either branch keeps behaving — the parity check the design doc calls out ("flag-on numbers must equal flag-off numbers") is 100% manual.

Not a blocker (no unit-test infra exists for `common-util/api/predict/*` today), but worth calling out: after the flag flips, a future refactor that hardcodes it on/off, deletes an `else`, or inlines the condition wouldn't fail anything. The pre-cut-over dry-run in the design's Verification section is load-bearing, and there's no safety net for regressions later.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,244 @@
...
+const QMR_WINDOW_SECONDS = QMR_MAX_AGE_DAYS * 86400;
+
+/** Fields we use from `/v1/data/scored-rows` rows. Other fields are ignored. */
+export type ScoredRow = {
+  request_id: string;
+  requester: string | null;
+  requested_at: string; // ISO 8601, tz-aware
+  question_title: string | null;
+  tool: string | null;
+  resolution_status: string;
```
DIvyaNautiyal07, line 35: **[P2]** `resolution_status: string` is compared as a literal at line 160 (`=== 'invalid'`). A `'pending' | 'resolved' | 'invalid'` union gives compile-time catches for typos and surfaces new endpoint values instead of silently letting them fall through. Same idea for the repeated `chain: 'gnosis' | 'polygon'` at every helper signature — export `MechAnalyticsChain = keyof typeof MECH_ANALYTICS_CHAIN_IDS` so adding a chain to the constant widens all four signatures automatically.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,244 @@
...
+  additions: Record<string, Record<string, number[]>>;
+  /** When true, the caller must drop the existing QMR map: `additions` holds the complete open set. */
+  rebuild: boolean;
+  /** New computed_at watermark to save. */
+  lastComputedAt: string;
+  /** Ids of rows we already processed (request_id → requested_at, unix seconds). */
+  ingestedRequestIds: Record<string, number>;
+  /** Highest requested_at we saved. Stored as lastMechRequestTimestamp so a rollback to the subgraph path can continue from it. */
+  lastTimestamp: number;
+  ok: boolean;
```
DIvyaNautiyal07: **[P2]** `{rebuild: true, ok: false}` is representable in this type but never returned by `fetchMechRequestsFromAnalytics` (that path becomes `null` at line 188). The caller at `roi-distribution.ts:476` has to re-derive intent from `result !== null && result.ok`.

A discriminated union makes the impossible state unrepresentable and the caller's branches exhaustive:
```ts
type RebuildResult = { kind: 'rebuild'; additions: ...; lastComputedAt: string; ... };
type IncrementalResult = { kind: 'incremental'; ok: boolean; additions: ...; lastComputedAt: string; ... };
export type AnalyticsQmrUpdate = RebuildResult | IncrementalResult;
```
The `null` return from the fetch stays as-is.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,244 @@
...
+    lastTimestamp,
+    ok,
+  };
+};
+
+/**
+ * Loads all scored rows for one requester made after the given time
+ * (unix seconds). Keeps no state between runs.
+ * `since` filters on requested_at; +1 second turns ">=" into ">".
+ */
```
DIvyaNautiyal07, line 246: **[P2]** `fetchScoredRowsForRequester` throws on any page failure (via `iterateScoredRows`), while its sibling `fetchMechRequestsFromAnalytics` catches. The asymmetry is deliberate — the current caller uses `Promise.allSettled` at `tool-accuracy.ts:206, 331` — but nothing in the signature or JSDoc says so. A future caller that doesn't wrap in `allSettled` gets an unhandled rejection. One-line note in the JSDoc would prevent that.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,244 @@
...
+  // watermark, so the next run also sees rows scored while we were fetching.
+  const runStart = new Date().toISOString();
+
+  const additions: Record<string, Record<string, number[]>> = {};
+  const ingestedRequestIds: Record<string, number> = isRebuild
+    ? {}
+    : { ...(previousIngestedIds ?? {}) };
+  let lastTimestamp = previousLastTimestamp;
+  // Updated row by row. See the AnalyticsQmrUpdate doc for why a partially
+  // advanced watermark is safe.
```
DIvyaNautiyal07: **[P2]** The watermark advances at line 146 *before* the duplicate/out-of-window/invalid checks. That's correct as designed (`since_computed_at` is inclusive, and `ingestedRequestIds` filters re-serves), but the comment on L126 doesn't say the design is load-bearing on the ID map. If someone later drops the map thinking it's overhead, silent double-count on every re-served row. Worth adding an explicit "safe only because `ingestedRequestIds` is always consulted first" line.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,244 @@
...
+      for (const row of rows) {
+        if (row.computed_at && (!maxComputedAt || row.computed_at > maxComputedAt)) {
+          maxComputedAt = row.computed_at;
+        }
+
+        if (ingestedRequestIds[row.request_id] !== undefined) {
+          skips.duplicate++;
+          continue;
+        }
+        const ts = Math.floor(Date.parse(row.requested_at) / 1000);
```
DIvyaNautiyal07: **[Nit]** `Math.floor(Date.parse(row.requested_at) / 1000)` also appears at `tool-accuracy.ts:104` in `fetchMechRequestsForSenderFromAnalytics`. Small `isoToUnixSec(iso: string): number` helper here (with the NaN guard from the P1 above folded in) would centralise the arithmetic and the validation.

common-util/api/predict/roi-distribution.ts
```diff
@@ -628,7 +675,11 @@ const updateAgentBlueprintData = async (
 
   return {
     mainData: { byDay, lastDayTimestamp, allTimeAgents, fetchErrors: runFetchErrors },
-    qmrData: { questionMechRequests: qmr, lastMechRequestTimestamp: newMechTs },
+    qmrData: {
+      questionMechRequests: qmr,
+      lastMechRequestTimestamp: newMechTs,
+      ...(newMechAnalytics !== undefined ? { mechAnalytics: newMechAnalytics } : {}),
```
DIvyaNautiyal07: **[Nit]** Since `mechAnalytics` is typed `?`, `mechAnalytics: newMechAnalytics` (which may be `undefined`) serializes identically to the conditional spread — JSON drops the key either way. The conditional spread can just be `mechAnalytics: newMechAnalytics,`.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,244 @@
...
+  const url = new URL(`${base.replace(/\/$/, '')}/v1/data/scored-rows`);
+  for (const [key, value] of Object.entries(searchParams)) {
+    url.searchParams.set(key, value);
+  }
+
+  let cursor: string | null = null;
+  while (true) {
+    if (cursor) url.searchParams.set('cursor', cursor);
+
+    const response = await fetch(url.toString());
```
DIvyaNautiyal07: **[Nit]** `fetch` accepts a `URL` object directly — the `.toString()` is redundant.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Guard added in a2f415ab — a computed_at value that does not parse can no longer become the watermark. One correction on the impact: the cursor could not pin permanently even before this. The round-1 watermark validation (`hasValidWatermark` on load) turns a bad persisted value into a warn + rebuild on the next run. So the worst case was one lost run, and now it is zero.

common-util/api/predict/roi-distribution.ts
Tanya-atatakai: Good catch — fixed in a2f415ab. A rebuild that receives zero rows now logs loudly and returns `null`, so the old QMR map stays and the next run retries. A rebuild with rows but an empty open set stays valid: received rows prove the endpoint has data.

common-util/api/predict/tool-accuracy.ts
Tanya-atatakai: Confirmed with a live query against the polygon marketplace subgraph, using the exact shape of `getMechRequestsBySenderWithToolQuery` (top-level `requests(where: { sender, blockTimestamp_gt })`): the top sender there has 2519 requests, the query returned rows with both `tool` and `questionTitle` populated, no errors. Both chains deploy the same `schema.graphql`, so the flag-off polygon path works. No code change.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Agreed, and no change in this PR — the repo has no test runner at all (see CLAUDE.md, Testing Considerations), so this cannot be solved locally for one module. The dry-run and source-comparison steps in docs/mech-analytics-migration.md are the gate for the flip, and flag removal later deletes one of the two branches anyway.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Partly fixed in a2f415ab — `MechAnalyticsChain = keyof typeof MECH_ANALYTICS_CHAIN_IDS` is exported and used in every signature in both files. Declined for `resolution_status`: the server owns that enum and can add values at any time; a client-side literal union would give false safety on a wire field. It stays `string`.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in a2f415ab — `AnalyticsQmrUpdate` is now `(Data & { kind: 'rebuild' }) | (Data & { kind: 'incremental'; ok: boolean })`, and `null` stays for a failed rebuild. `{rebuild: true, ok: false}` is no longer representable, and the caller branches on `kind`.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in a2f415ab — the JSDoc now states the function throws on any page failure and that callers must handle the rejection.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in a2f415ab — the comment at the watermark update now states the order is safe only because `ingestedRequestIds` is always checked before a row is ingested.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in a2f415ab — `isoToUnixSec` is exported from the client and used in both files. The NaN guard stays at the call sites, because only the QMR path needs to count the skip.

common-util/api/predict/roi-distribution.ts
Tanya-atatakai: Fixed in a2f415ab — plain `mechAnalytics: newMechAnalytics`.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in a2f415ab.

common-util/api/predict/tool-accuracy.ts
```diff
@@ -81,15 +89,39 @@ async function fetchResolvedBets(): Promise<Bet[]> {
...
+  chain: MechAnalyticsChain,
   sender: string,
   timestampGt: number
 ): Promise<MechRequest[]> {
   const allRequests: MechRequest[] = [];
 
   for (let i = 0; i < MECH_MAX_PAGES; i++) {
     const skip = i * MECH_PAGE_SIZE;
-    const data = await MARKETPLACE_GRAPH_CLIENTS.gnosis.request<{
+    const data = await MARKETPLACE_GRAPH_CLIENTS[chain].request<{
```
DIvyaNautiyal07, line 124: **[P1] Silent query-shape change on the polygon flag-off path.** The removed `fetchPolygonMechRequestsForSender` queried `sender(id: "...") { requests(where: { blockTimestamp_gt }) }`. This shared function now uses `getMechRequestsBySenderWithToolQuery`, which hits the top-level `requests(where: { sender: ..., blockTimestamp_gt: ... })` entity for both chains. The polygon marketplace subgraph does expose the top-level `requests` entity (roi-distribution's incremental query uses it), but the `where: { sender: ... }` filter on it has never been exercised on polygon in production. If `sender` isn't indexed the same way as on gnosis, per-sender pulls silently return `[]` and the outer `Promise.allSettled` swallows the empty result. Please confirm on a preview with `USE_MECH_ANALYTICS=false` that `computePolystratToolAccuracy` still returns non-empty per-sender rows before merging.

common-util/api/predict/tool-accuracy.ts
```diff
@@ -164,7 +208,7 @@ export async function computeOmenstratToolAccuracy(): Promise<ToolAccuracyStat[]
   for (let i = 0; i < senders.length; i += SENDER_BATCH_SIZE) {
     const batch = senders.slice(i, i + SENDER_BATCH_SIZE);
     const results = await Promise.allSettled(
-      batch.map((sender) => fetchMechRequestsForSender(sender, earliestBetTimestamp))
+      batch.map((sender) => fetchMechRequestsForSender('gnosis', sender, earliestBetTimestamp))
```
DIvyaNautiyal07, line 211: **[P1] Rejected batch items are silently dropped (both this loop and the polystrat mirror around line 336).** The subsequent `forEach` only handles `fulfilled` — a rejection is not logged, not counted, and the bettor is silently excluded from `mechRequestsBySender` (`.get(bettorId) ?? []` → `continue`). This was already the shape for polygon, but the risk profile changed: with the flag on and `MECH_ANALYTICS_URL` unset, `iterateScoredRows` throws on every call and the snapshot is saved as `{ omenstrat: [], polystrat: [] }` looking healthy.

Fix:
```ts
results.forEach((result, idx) => {
  if (result.status === 'fulfilled') {
    mechRequestsBySender.set(batch[idx], result.value);
  } else {
    console.error(`[tool-accuracy] fetch failed for ${batch[idx]}`, result.reason);
  }
});
```

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,270 @@
...
+          skips.duplicate++;
+          continue;
+        }
+        const ts = isoToUnixSec(row.requested_at);
+        if (!Number.isFinite(ts) || ts <= 0 || ts < windowStartSec) {
+          skips.outOfWindow++;
+          continue;
+        }
+
+        if (isRebuild && row.resolution_status === 'invalid') {
```
DIvyaNautiyal07: **[P2] `resolution_status === 'invalid'` is filtered only on rebuild, not incremental.** The `isRebuild &&` guard means invalid rows re-served during an incremental run (after a sweep touches their `computed_at`) pass every check and get counted as open QMR entries, inflating `openRequestCount` slightly until the 14-day TTL flush. Bounded impact, but the two paths behave differently and the spec doesn't call it out. Either drop the `isRebuild &&` prefix or document the incremental drift.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,270 @@
...
+      throw new Error(`mech-analytics /v1/data/scored-rows returned ${response.status}`);
+    }
+    const page = (await response.json()) as ScoredRowsPage;
+    // A 200 without a rows array is a broken response, not an empty page.
+    if (!Array.isArray(page.rows)) {
+      throw new Error('mech-analytics /v1/data/scored-rows returned no rows array');
+    }
+    yield page.rows;
+
+    if (!page.next_cursor) break;
```
DIvyaNautiyal07, line 73: **[P2] `next_cursor` type is not validated.** The response is cast `as ScoredRowsPage` with no runtime shape check. If the API returns a non-string truthy `next_cursor`, the falsy break passes, `cursor = page.next_cursor` sets a non-string value, and `url.searchParams.set('cursor', cursor)` stringifies it (e.g. `[object Object]`). The API then either 400s (swallowed as `ok=false`) or the loop wastes pages on a garbage cursor. Cheap fix:
```ts
if (typeof page.next_cursor !== 'string' || !page.next_cursor) break;
```

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,270 @@
...
+    requester,
+    since: new Date((timestampGt + 1) * 1000).toISOString(),
+    limit: String(pageSize),
+  };
+  for await (const page of iterateScoredRows(params)) {
+    rows.push(...page);
+    lastPageFull = page.length === pageSize;
+    if (++pages >= maxPages) break;
+  }
+  if (pages >= maxPages && lastPageFull) {
```
DIvyaNautiyal07, line 270: **[P2] Cap-hit warning has a false-positive case.** If the API happens to return exactly `pageSize * maxPages` rows and the next page would have been empty (`next_cursor` about to be null), `lastPageFull=true` and `pages >= maxPages` are both satisfied and the warning fires despite no data being lost. Log-only, but likely to confuse oncall. Either surface `next_cursor` through `iterateScoredRows` so the truncation check is exact, or note the false-positive case in the comment above.

common-util/api/predict/mech-analytics.ts
```diff
@@ -0,0 +1,270 @@
...
+ * because the subgraph has no data for off-chain requests.
+ * Full details: docs/mech-analytics-migration.md
+ */
+
+/** One switch that turns on all mech-analytics read paths. */
+export const USE_MECH_ANALYTICS = process.env.USE_MECH_ANALYTICS === 'true';
+
+const PAGE_SIZE = 5000; // endpoint max limit
+
+export const MECH_ANALYTICS_CHAIN_IDS = {
```
DIvyaNautiyal07, line 15: **[P2] `MECH_ANALYTICS_CHAIN_IDS` keys are not constrained to a subset of `MARKETPLACE_GRAPH_CLIENTS` keys.** Both `gnosis` and `polygon` currently exist in `MARKETPLACE_GRAPH_CLIENTS` so today this is safe — but nothing type-enforces the coupling. Adding a chain here without a matching graph client would silently return `undefined` from `MARKETPLACE_GRAPH_CLIENTS[chain]` in `tool-accuracy.ts:124` and roi-distribution's incremental query. Optional: type `MARKETPLACE_GRAPH_CLIENTS` (or a derived object) as `Record<MechAnalyticsChain, GraphQLClient>` so the compiler catches drift.

common-util/api/predict/roi-distribution.ts
```diff
@@ -14,14 +14,16 @@ import {
...
 const LIMIT = 1000;
 // Process at most this many days per cron run to stay within timeout
 const MAX_DAYS_PER_RUN = 30;
 const DAY_SECONDS = 86400;
 // Keep only this many days in byDay (covers the longest non-global tab: 90D)
 const BYDAY_RETENTION_DAYS = 90;
-// FIX-2: age-out TTL for pending QMR entries (markets resolve in ~4 days)
-const QMR_MAX_AGE_DAYS = 14;
+// QMR_MAX_AGE_DAYS (common-util/constants.ts): age-out TTL for pending QMR
+// entries (markets resolve in ~4 days); shared with the mech-analytics client.
```
DIvyaNautiyal07, line 26: **[Nit] Comment block for `MIN_TRADES_FOR_ROI_DISPLAY` is fused with the `QMR_MAX_AGE_DAYS` note above.** Removing the local `QMR_MAX_AGE_DAYS` constant left its 2-line explanation glued directly to the next comment (about `MIN_TRADES_FOR_ROI_DISPLAY`) with no blank line. The two now read as one block about unrelated constants. Add a blank line after this line.

common-util/api/predict/roi-distribution.ts
```diff
@@ -445,11 +457,46 @@ const updateAgentBlueprintData = async (
...
+
+  if (USE_MECH_ANALYTICS) {
+    const result = await fetchMechRequestsFromAnalytics(
+      chain,
+      existingQmr?.mechAnalytics?.lastComputedAt,
+      existingQmr?.mechAnalytics?.ingestedRequestIds,
+      newMechTs
+    );
+    // null = the rebuild failed; keep everything as it is and retry next run.
+    mechRequestsOk = result !== null && (result.kind === 'rebuild' || result.ok);
```
DIvyaNautiyal07, line 477: **[Nit] `mechRequestsOk` declared uninitialized at line 462.** Under `strict: false` the compiler doesn't catch an uninitialized read, and today the `if/else` covers both branches. If a future refactor adds an early return in either branch, this reads uninitialized without a compiler warning. Safer as `let mechRequestsOk = false;` at the declaration.

common-util/api/predict/tool-accuracy.ts
Tanya-atatakai: This exact filter was exercised in the round-2 check (see my reply on the earlier polygon thread): I ran the top-level `requests(where: { sender, blockTimestamp_gt })` shape live against the polygon marketplace subgraph for its top sender (2519 requests) and got rows back with `tool` and `questionTitle` populated, no errors. So the `sender` where-filter works on polygon, not only the entity shape. No code change.

common-util/api/predict/tool-accuracy.ts
Tanya-atatakai: Good catch — fixed in 5b6eee08. Both `allSettled` loops now log a `console.error` per rejected sender, so an outage (e.g. missing `MECH_ANALYTICS_URL` with the flag on) is visible in the cron logs instead of looking like healthy empty data.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in 5b6eee08 — the `invalid` skip now applies on both paths, and the row ID is still recorded so later re-serves stay skipped. The spec's incremental skip rules are updated to match.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Fixed in 5b6eee08, slightly stricter than proposed: a truthy non-string `next_cursor` now throws (consistent with the missing-rows-array check) instead of silently breaking with partial data. `null` still ends the loop normally.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Noted in 5b6eee08 — a comment above the warning documents the exactly-pageSize×maxPages false positive. The exact check needs `next_cursor` surfaced through the generator API; not worth the churn for a log line.

common-util/api/predict/mech-analytics.ts
Tanya-atatakai: Declined. The coupling check needs the REST client to type-import from the GraphQL layer for a two-entry constant that almost never changes, and the new rejected-fetch logging (5b6eee08) surfaces exactly this failure at runtime. If the chain list ever grows, the spec's chain section is the checklist.

common-util/api/predict/roi-distribution.ts
Tanya-atatakai: Fixed in 5b6eee08.

common-util/api/predict/roi-distribution.ts
Tanya-atatakai: Fixed in 5b6eee08 — `let mechRequestsOk = false`.

DIvyaNautiyal07 (approved):
Round 3 (5b6eee08) addresses everything actionable: rejection logging in both `allSettled` loops, unconditional `invalid` skip on incremental with ID recording, non-string `next_cursor` throws, cap-warning false positive documented, `mechRequestsOk` initialized, comment spacing. The polygon `where: { sender }` filter was verified live against the marketplace subgraph (2519 rows returned with `tool`/`questionTitle` populated), so the P1 verification is satisfied. The type-coupling nit was declined with reasonable justification. LGTM.

atepem (approved):

PR #551, opened 2026-08-05
Add mech-analytics parity verification script (feed-swap gate)

DIvyaNautiyal07:
## Summary

Parity gate for the olas-website → mech-analytics migration, scoped to what actually migrates per `mech-analytics/docs/consumer_migration.md` §4 + §12 Q4: **the per-request mech-request feed swap** behind `USE_MECH_ANALYTICS` (PR #548) — `fetchIncrementalMechRequests` → `/v1/data/scored-rows`, `senderTotal` → `agent_aggregates.n_mech_requests`. partialRoi/successRate keep their site-side formulas and are out of scope.

**The gate** (`scripts/verify-mech-analytics-parity.mjs`, `yarn verify-mech-analytics-parity`), per platform (omenstrat = gnosis/100/14, polystrat = polygon/137/86), last-4-days window with a lag buffer for scoring delay:
1. **Open-market count parity (core):** scored-rows `pending` + `market_id` count vs the site's open count (marketplace-subgraph feed minus settlement consumption via normalizeTitle matching, mirroring `roi-distribution.ts`). Reports the (a) pending+market_id / (b) invalid / (c) no-market_id breakout. Exact.
2. **Row parity:** multiset diff on (requester, unix ts, normalized title) between `getMechRequestsIncrementalQuery` rows and scored-rows; missing/extra reported with samples. Exact.
3. **senderTotal parity:** `windows.all.n_mech_requests` vs Σ(totalLegacy + totalMarketplace) over the Safes from `/instances`. ±5 absolute (settlement-lag/rollup-cache allowance, configurable).

The flag-on vs flag-off `openRequestCount` check runs in the ops staging parallel-run (needs the Next/Blob runtime) and is documented in the script header.

**Exit codes:** 0 PASS · 1 unexpected error (sentinel) · 2 all-vacuous (nothing comparable never reads as a pass) · 3 divergence · 4 data gap (e.g. subgraph skip-cap truncation). Divergence outranks gap. Null is never coerced to 0. `finally` always writes `parity-<ts>.md` + `.json` artifacts. Subgraph URLs embed API keys — error output is scrubbed.

**Docker/CI:** `Dockerfile.verify` (node:22-slim, non-root, GIT_SHA baked, node-builtins only so no dependency install layer) + `publish-verify-image.yaml` (publish on release as `valory/olas-website-verify:<version>`; fork-safe build-only + `--help` smoke on PRs touching image inputs).

Read-only and additive — no production code paths touched.

## Test plan

- [x] `yarn lint` clean
- [x] All five exit paths proven against a local mock (pass=0, divergence=3, vacuous=2, gap=4, gap+divergence=3); artifacts written
- [x] Docker build; `--help` exit 0; missing-env run exits 1 with a clean config-error message
- [ ] Production team: real run with the 4 subgraph URLs + `MECH_ANALYTICS_URL`; post the PASS artifact here before cutover

### DIFF
--- a/.github/workflows/publish-verify-image.yaml
+++ b/.github/workflows/publish-verify-image.yaml
@@ -6,13 +6,15 @@ name: Publish verify image
 # scripts/verify-mech-analytics-parity.mjs).
 #
 # * On release: builds AND pushes tagged image to Docker Hub.
-# * On PR touching Dockerfile.verify, the verify script, the workflow,
-#   or the deps that end up in the image: builds only (no push, no
-#   credentials needed — safe on fork PRs) and runs ``--help`` against
-#   the built image so the entry point wiring is exercised end-to-end.
-#   Prevents shipping a broken Dockerfile (chown ordering, module
-#   resolution, deps drift) whose only signal otherwise is "infra tries
-#   to run it and it crashes".
+# * On PR touching Dockerfile.verify, the verify script, or the
+#   workflow: builds only (no push, no credentials needed — safe on
+#   fork PRs) and runs ``--help`` against the built image so the entry
+#   point wiring is exercised end-to-end. Prevents shipping a broken
+#   Dockerfile (chown ordering, entrypoint wiring) whose only signal
+#   otherwise is "infra tries to run it and it crashes". The image
+#   installs no npm dependencies (the script is node built-ins only),
+#   so package.json / yarn.lock changes cannot affect it and are not
+#   triggers.
 
 on:
   release:
@@ -21,9 +23,6 @@ on:
     paths:
       - 'Dockerfile.verify'
       - 'scripts/verify-mech-analytics-parity.mjs'
-      - 'package.json'
-      - 'yarn.lock'
-      - '.yarnrc'
       - '.github/workflows/publish-verify-image.yaml'
 
 # Least-privilege default for GITHUB_TOKEN (same discipline as main.yml):
--- a/.github/workflows/publish-verify-image.yaml
+++ b/.github/workflows/publish-verify-image.yaml
@@ -1,0 +1,102 @@
+name: Publish verify image
+
+# Builds and pushes valory/olas-website-verify — the image infra runs as
+# a K8s Job for the predict-page / mech-analytics parity verification
+# (Dockerfile.verify, entry point is
+# scripts/verify-mech-analytics-parity.mjs).
+#
+# * On release: builds AND pushes tagged image to Docker Hub.
+# * On PR touching Dockerfile.verify, the verify script, the workflow,
+#   or the deps that end up in the image: builds only (no push, no
+#   credentials needed — safe on fork PRs) and runs ``--help`` against
+#   the built image so the entry point wiring is exercised end-to-end.
+#   Prevents shipping a broken Dockerfile (chown ordering, module
+#   resolution, deps drift) whose only signal otherwise is "infra tries
+#   to run it and it crashes".
+
+on:
+  release:
+    types: [prereleased, released]
+  pull_request:
+    paths:
+      - 'Dockerfile.verify'
+      - 'scripts/verify-mech-analytics-parity.mjs'
+      - 'package.json'
+      - 'yarn.lock'
+      - '.yarnrc'
+      - '.github/workflows/publish-verify-image.yaml'
+
+# Least-privilege default for GITHUB_TOKEN (same discipline as main.yml):
+# both jobs only need to read the repo. Docker Hub pushes authenticate
+# with repo secrets, not GITHUB_TOKEN.
+permissions:
+  contents: read
+
+jobs:
+  publish:
+    if: github.event_name == 'release'
+    runs-on: ubuntu-latest
+    env:
+      IMAGE: valory/olas-website-verify
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Compute image tag
+        id: tag
+        run: |
+          tag="${GITHUB_REF_NAME#v}"
+          echo "tag=$tag" >> "$GITHUB_OUTPUT"
+          echo "Computed tag: $tag"
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Build and push
+        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6
+        with:
+          context: .
+          file: Dockerfile.verify
+          push: true
+          tags: ${{ env.IMAGE }}:${{ steps.tag.outputs.tag }}
+          # Bake the source commit into the image so the verify script's
+          # metadata header can attribute each artifact to a specific
+          # revision without needing git inside the runtime image.
+          build-args: |
+            GIT_SHA=${{ github.sha }}
+
+  build-only:
+    # PR-triggered smoke build — no registry push, no secrets required,
+    # so it runs safely on fork PRs. Load the image into the local
+    # Docker daemon and invoke ``--help`` to catch any regression in
+    # Dockerfile layer ordering, dependency install, or entrypoint that
+    # would otherwise only surface when infra tries to run the released
+    # tag.
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    env:
+      IMAGE: valory/olas-website-verify:pr-${{ github.event.pull_request.number }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
+
+      - name: Build (no push)
+        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6
+        with:
+          context: .
+          file: Dockerfile.verify
+          push: false
+          load: true
+          tags: ${{ env.IMAGE }}
+          build-args: |
+            GIT_SHA=${{ github.sha }}
+
+      - name: Smoke-test entrypoint
+        run: docker run --rm ${{ env.IMAGE }} --help
--- a/scripts/verify-mech-analytics-parity.mjs
+++ b/scripts/verify-mech-analytics-parity.mjs
@@ -107,6 +107,15 @@ const LAG_BUFFER_MINUTES_DEFAULT = 30;
 // are otherwise compared exactly (checks 1 and 2 use zero tolerance).
 const COUNT_TOLERANCE_DEFAULT = 5;
 
+// Mech-analytics keeps a 24h finality gate on Omen: rows only flip to
+// `resolved` 24h after finalization (consumer_migration.md §6). The
+// subgraph's dailyProfitStatistics consumes a market as soon as the
+// day's stats record it, so markets resolved in the last ~24h are
+// consumed on the subgraph side but still `pending` in analytics.
+// Over a 4-day window that's a chronic delta that turns honest runs
+// red. Cap check-1 dailyStats consumption at (now - N h) to align.
+const FINALITY_LAG_HOURS_DEFAULT = 24;
+
 const DAY_SECONDS = 86400;
 const SUBGRAPH_PAGE = 1000;
 // graph-node rejects skip > 5000; a full page at that offset means the
@@ -158,6 +167,7 @@ const { values: args } = parseArgs({
     'output-dir': { type: 'string' },
     'window-days': { type: 'string' },
     'lag-buffer-minutes': { type: 'string' },
+    'finality-lag-hours': { type: 'string' },
     'count-tolerance': { type: 'string' },
     verbose: { type: 'boolean', short: 'v' },
     help: { type: 'boolean', short: 'h' },
@@ -186,6 +196,8 @@ if (args.help) {
       '  --mech-analytics-url <url>   mech-analytics base URL (or MECH_ANALYTICS_URL env)\n' +
       `  --window-days <n>            checks 1+2 window (default ${WINDOW_DAYS_DEFAULT})\n` +
       `  --lag-buffer-minutes <n>     near-edge trim for scoring lag (default ${LAG_BUFFER_MINUTES_DEFAULT})\n` +
+      `  --finality-lag-hours <n>     check 1 consumption cutoff for the analytics 24h\n` +
+      `                               finality gate (default ${FINALITY_LAG_HOURS_DEFAULT})\n` +
       `  --count-tolerance <n>        check 3 absolute tolerance (default ${COUNT_TOLERANCE_DEFAULT})\n` +
       '  --output-dir <dir>           write parity-<ts>.md/.json artifacts here\n' +
       '  -v, --verbose                log per-page fetch progress\n' +
@@ -234,6 +246,11 @@ const lagBufferMinutes = parseNonNegativeInt(
   args['lag-buffer-minutes'],
   'lag-buffer-minutes',
   LAG_BUFFER_MINUTES_DEFAULT
+);
+const finalityLagHours = parseNonNegativeInt(
+  args['finality-lag-hours'],
+  'finality-lag-hours',
+  FINALITY_LAG_HOURS_DEFAULT
 );
 const countTolerance = parseNonNegativeInt(
   args['count-tolerance'],
@@ -298,6 +315,7 @@ const metadata = {
   mech_analytics_url: safeOrigin(mechAnalyticsUrl),
   window_days: windowDays,
   lag_buffer_minutes: lagBufferMinutes,
+  finality_lag_hours: finalityLagHours,
   count_tolerance_abs: countTolerance,
 };
 
@@ -307,6 +325,7 @@ emit(`  script git SHA:       ${metadata.script_git_sha}`);
 emit(`  mech-analytics URL:   ${metadata.mech_analytics_url}`);
 emit(`  window (days):        ${metadata.window_days}`);
 emit(`  lag buffer (min):     ${metadata.lag_buffer_minutes}`);
+emit(`  finality lag (h):     ${metadata.finality_lag_hours} (check 1 only)`);
 emit(`  count tolerance (±):  ${metadata.count_tolerance_abs} (check 3 only)`);
 for (const name of REQUIRED_ENV.slice(1)) {
   // Presence only — the URLs embed API keys.
@@ -373,6 +392,7 @@ const fetchMarketplaceRequests = async (url, windowStartSec, windowEndSec, label
           orderBy: blockTimestamp
           orderDirection: asc
         ) {
+          id
           sender { id }
           blockTimestamp
           parsedRequest { questionTitle }
@@ -408,7 +428,15 @@ const fetchMarketplaceRequests = async (url, windowStartSec, windowEndSec, label
       skippedAfterWindowEnd += 1;
       continue;
     }
-    usable.push({ requester, ts, title });
+    // req.id is the marketplace subgraph's Request.id — same identifier
+    // the mech-analytics lake writes back as `request_id`, so check 2
+    // can match on it when both sides have it (see diffRowSets).
+    usable.push({
+      requester,
+      ts,
+      title,
+      requestId: typeof req.id === 'string' ? req.id.toLowerCase() : null,
+    });
   }
   return { rows: usable, truncated, skippedMissingKey, skippedAfterWindowEnd };
 };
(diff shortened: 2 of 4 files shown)


scripts/verify-mech-analytics-parity.mjs
```diff
@@ -0,0 +1,449 @@
+// Copyright 2026 Valory AG
+// SPDX-License-Identifier: Apache-2.0
+//
+// Parity verification: the olas-website predict page's `partialRoi` and
+// `successRate` fields are the two metrics migrating to mech-analytics
+// per the consumer-migration plan (see mech-analytics docs). This
+// script proves that switching those two fields from the current
+// subgraph-derived computation to the mech-analytics API preserves the
+// numbers a visitor sees. Every other field on the predict page keeps
+// its current computation and isn't the script's concern.
```
Tanya-atatakai: please let's align this with the consumer-migration plan before merging — per `consumer_migration.md` §4 and olas-website #548, `partialRoi` and `successRate` are NOT migrating to the ai-agent endpoint. the website keeps its own formulas and only swaps the per-request mech feed to `/v1/data/scored-rows`. so this compares two independent implementations with at least four known definitional gaps: (1) population — our partialRoi aggregates every trader agent in the roi-distribution blob, `100/14`+`100/25` covers only registered Safes of those two blueprints; (2) successRate is "% of settled bets correct" (accuracy.ts), `tool_accuracy_*` is mech-prediction accuracy — a different metric by definition; (3) our costs include mech fees at `DEFAULT_MECH_FEE`, `roi_omen` doesn't; (4) the settlement-vs-placement bucketing you already name. with rel tolerance 1e-4 this gate exits 3 on every honest run. the check the doc prescribes (§12 Q4) is `openRequestCount` flag-on vs flag-off + scored-rows row parity. is this meant as an informational cross-check (then it shouldn't gate with exit 3), or the go/no-go (then it gates the wrong thing)? please confirm the intended pass condition with the mech-analytics owner 🙏

Tanya-atatakai (commented):
reviewed with the consumer-migration docs side by side — the script itself is carefully built (sentinel discipline, artifact writing, null handling are all solid), but I have one question about the premise that decides everything else, plus a few concrete fixes. verdict: needs changes, mainly pending the premise answer.

**outside this PR (no action here):** the chain/agent constants (100/14, 100/25, 137/86) now live in three repos — worth a ticket for a shared source once the migration settles. and the checks `consumer_migration.md` §12 Q4 actually prescribes for olas-website (openRequestCount flag-on vs flag-off, scored-rows row parity) need a home — either this script grows them or a sibling does; worth a ticket either way.

**what I verified:** probed staging `/v1/metrics/ai-agent/100/14` (window keys, field names, `total_cost/payout_*_usd` presence); confirmed `predict-v2` + schema `20260708` against `snapshot-storage.ts`; confirmed `successRate` = bet win-rate by placement day (`common-util/api/predict/index.ts:233`); cross-checked the migration claims against `consumer_migration.md` §4/§12.

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -0,0 +1,449 @@
...
+//
+// mech-analytics itself aggregates multi-Safe agents in
+// _aggregate_agent (etl/api/routes/agent.py:303). It uses:
+//   * ROI:  sum(payout) / sum(cost)  — never mean of ratios
+//   * Accuracy: bet-count-weighted mean of per-Safe accuracy
+// Because the per-agent endpoint returns already-aggregated ROI (not
+// raw payout/cost), a client-side sum can't perfectly reproduce
+// mech-analytics's payout-weighted ROI for the omenstrat pair. We
+// approximate with an n_bets-weighted mean of ROI, which is the
+// best we can do from the exposed fields, and note the caveat in
```
Tanya-atatakai: "the per-agent endpoint returns already-aggregated ROI (not raw payout/cost)" — this isn't true anymore: the window block exposes `total_cost_omen_usd` and `total_payout_omen_usd` (verified on staging `/v1/metrics/ai-agent/100/14`). so the exact payout-weighted aggregation IS computable: `sum(payout) / sum(cost)` across 14+25, same as `_aggregate_agent`. can we do that instead of the n_bets-weighted mean-of-ratios and drop the approximation caveat entirely?

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -0,0 +1,449 @@
...
+
+const compareCell = (oldValue, newValue) => {
+  if (oldValue == null && newValue == null) return { status: 'both-null' };
+  if (oldValue == null || newValue == null) {
+    return { status: 'null-mismatch', oldValue, newValue };
+  }
+  if (!Number.isFinite(oldValue) || !Number.isFinite(newValue)) {
+    return { status: 'non-finite', oldValue, newValue };
+  }
+  const denom = Math.max(Math.abs(oldValue), 1e-6);
```
Tanya-atatakai: the relative-only tolerance blows up near zero: ROI can legitimately sit at ~0%, and `denom = max(|old|, 1e-6)` turns a 0.001pp float wobble into a FAIL. can we add an absolute floor for percent cells — e.g. pass when `absDiff <= 0.01` (one basis point) OR rel within tolerance? the townhall sibling already uses absolute mode for fees.

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -0,0 +1,449 @@
...
+    (r) => r.status === 'fail' || r.status === 'null-mismatch' || r.status === 'non-finite'
+  );
+  const missing = cellResults.filter((r) => r.status === 'missing-field');
+  emit(`  cells checked:       ${cellResults.length}`);
+  emit(`  cells passed:        ${cellResults.filter((r) => r.status === 'pass').length}`);
+  emit(`  cells both-null:     ${cellResults.filter((r) => r.status === 'both-null').length}`);
+  emit(`  cells failed:        ${failures.length}`);
+  emit(`  cells missing field: ${missing.length}`);
+
+  if (missing.length > 0) {
```
Tanya-atatakai: if every cell is `both-null` (blob written by a failing run + empty analytics), the verdict is PASS exit 0 — a vacuous pass. the townhall sibling has exit 2 for exactly this; can we add the same guard here? also jsyk the exit precedence is inverted vs the sibling: here missing-field(4) outranks divergence(3), there divergence outranks gap with a written rationale ("a gap must not mask a regression"). I'd rather both scripts agree, and the sibling's order looks right to me.

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -0,0 +1,449 @@
...
+let exitCode = 1; // sentinel: any unexpected exit stays 1
+let verdict = 'ERROR';
+const cellResults = [];
+
+try {
+  section('Data pulls');
+  emit('  pulling olas-website snapshot from Vercel Blob...');
+  const olasData = await fetchOlasWebsiteSnapshot();
+  emit(
+    `  snapshot fetched (timestamp: ${olasData.omenstrat?.partialRoi?.status?.checkedAt || 'unknown'})`
```
Tanya-atatakai: tiny, `status.checkedAt` doesn't exist on `MetricStatus` — the field is `lastValidAt` — so this always prints "unknown":
```suggestion
    `  snapshot fetched (timestamp: ${olasData.omenstrat?.partialRoi?.status?.lastValidAt || 'unknown'})`
```

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -0,0 +1,449 @@
...
+// ---------------------------------------------------------------------------
+
+const SNAPSHOT_CATEGORY = 'predict-v2';
+
+// Same versioning discipline snapshot-storage.ts uses. Bump when the
+// upstream constant changes.
+const SCHEMA_VERSION = '20260708';
+
+const OMENSTRAT_AGENTS = [
+  { chainId: 100, agentId: 14 },
```
Tanya-atatakai: where does agent 25 come from? the migration doc pins omenstrat to `100/14` in §4, and its own note on `TRADER_AGENT_IDS` says the 14+25 cite "may be stale — verify against olas-operate-app". can you confirm 25 with the trader team, so the pair is a verified fact and not an inherited constant?

.github/workflows/publish-verify-image.yaml
```diff
@@ -0,0 +1,102 @@
...
+          echo "Computed tag: $tag"
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
```
atepem, line 57: `DOCKER_PASSWORD` isn't available to this repo, so this job can't succeed as written. Checking the API: repo-level Actions secrets are empty (`total_count: 0`) and the only org secret exposed here is `DOCKER_USER`. `docker/login-action` gets an empty password and the `publish` job fails at this step on the first release.

It's latent right now — the repo has no releases yet, and the PR-triggered `build-only` job needs no credentials, which is why CI is green — but this ships a workflow whose push path has never run and currently can't. Could infra add the secret, or should the `publish` job be dropped until the image is actually consumed?

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -0,0 +1,449 @@
...
+if (!mechAnalyticsUrl) {
+  console.error('MECH_ANALYTICS_URL (env or --mech-analytics-url) is required');
+  process.exit(1);
+}
+if (!process.env.BLOB_READ_WRITE_TOKEN) {
+  console.error('BLOB_READ_WRITE_TOKEN is required to read the olas-website snapshot');
+  process.exit(1);
+}
+const outputDir = args['output-dir'] ? path.resolve(args['output-dir']) : null;
+const tolerance = args.tolerance ? Number(args.tolerance) : DEFAULT_TOLERANCE_REL;
```
atepem: A malformed `--tolerance` silently fails every cell instead of erroring. `Number('abc')` is `NaN`, and `relDiff <= NaN` is always false, so all 16 cells come back `fail` and the script exits 3 — indistinguishable from a real divergence, with nothing in the report pointing at the bad flag.

Worth a guard given exit 3 is the signal someone acts on:

```suggestion
const tolerance = args.tolerance ? Number(args.tolerance) : DEFAULT_TOLERANCE_REL;
if (!Number.isFinite(tolerance) || tolerance < 0) {
  console.error(`--tolerance must be a non-negative number, got: ${args.tolerance}`);
  process.exit(1);
}
```

(Exit 1 rather than 3 keeps it in the "operator error" bucket, consistent with the missing-env checks above.)

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -0,0 +1,449 @@
...
+  '7d': '7d',
+  '30d': '30d',
+  '90d': '90d',
+  max: 'all',
+};
+
+// Default tolerance: relative 1e-4 for percent-valued metrics. Percent
+// scale means a 0.01% delta trips this; anything smaller is
+// float-precision noise. Both sides pull integer counts through
+// float arithmetic so exact equality is not a realistic bar.
```
atepem: Small factual fix while this block is being revisited for the near-zero floor: the comment's arithmetic doesn't match the constant. `1e-4` is relative to the compared value, not an absolute percentage-point threshold — on a ROI of `12.5` it trips at a gap of `0.00125`pp, roughly 8x stricter than the "0.01% delta" the comment describes. Worth correcting so nobody re-tunes the threshold from the wrong number.

atepem (commented):
Supplementing @Tanya-atatakai's review rather than repeating it — I checked their premise question against `docs/mech-analytics-migration.md` in this repo and it holds up: the "What migrates" table lists only the ROI incremental mech-request feed and the tool-accuracy per-sender pulls, both to `/v1/data/scored-rows`, and the Verification section prescribes `openRequestCount` parity flag-on vs flag-off. `partialRoi` / `successRate` aren't in either list. So I'd resolve that thread before spending time on the mechanics below.

Three things their review doesn't cover. The first is the only one I'd call blocking for the workflow to be functional at all; the other two are small.

Rest of the change is genuinely low-risk from the website's side — no `pages/`, `components/`, or `common-util/` file is touched, and I verified the script's constants against the repo (`predict-v2` matches `PREDICT_SNAPSHOT_CATEGORY`, `20260708` matches `DEFAULT_SCHEMA_VERSION` with an empty `SCHEMA_VERSIONS`, window keys match `WindowKey`, blob filename and `snapshot.data` shape match `getSnapshot`, `@vercel/blob` is a prod dep so `--production` resolves it).

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: You're right on the premise. Confirmed with the migration owner: partialRoi and successRate are not moving to the ai-agent endpoint. The website keeps its own formulas and the migration only swaps the per-request feed to /v1/data/scored-rows, per consumer_migration.md §4. So this script was gating the wrong thing, and you're also right it would exit 3 on every honest run.

Reworking it now to the §12 Q4 checks: open-market count parity (pending + market_id vs the normalizeTitle-based count, with the invalid and no-market_id breakouts reported), scored-rows row parity against the marketplace-subgraph feed, and senderTotal vs n_mech_requests. The reframe drops the blob snapshot and the omenstrat aggregation entirely, which moots the lastValidAt and agent-25 points. Also taking the vacuous-pass guard (exit 2), flipping the exit precedence so divergence outranks gap, and adding an absolute floor next to the relative tolerance. Will push the rework to this PR.

.github/workflows/publish-verify-image.yaml
DIvyaNautiyal07: Confirmed, repo secrets are empty and only DOCKER_USER is exposed from the org, so the publish path can't run today. Since the image is meant to be run by the production team for the cutover gate, I'd keep the publish job and have infra add DOCKER_PASSWORD before cutting the first release, the PR-triggered build-only job stays the CI gate meanwhile. If you'd rather not ship a never-run push path I'm fine dropping the publish job until the image is actually consumed, say the word.

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: The rework replaced this with parseNonNegativeInt for --count-tolerance, which does what you suggest: Number('abc') fails the Number.isInteger check and exits 1 with an operator-error message, so a malformed flag can't masquerade as exit 3 anymore.

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Mooted by the rework, the relative float tolerance is gone entirely. Counts are compared exactly and the only tolerance left is the integer absolute slack on the senderTotal check, documented as a settlement-lag allowance.

atepem (dismissed):

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Mooted by the rework in 5b4f1f14 — the omenstrat aggregation is gone entirely because partialRoi doesn't migrate anymore, so mean-of-ratios vs payout-weighted is a non-issue in this script. Good catch on total_cost_omen_usd / total_payout_omen_usd though: filing that away for the mech-predict rewrite (analyze.py on top of /v1/data/scored-rows), where the same aggregation problem does reappear.

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Mooted by the rework — there are no float compares left. The three checks are all count parity: checks 1 and 2 are exact, check 3 uses an absolute integer tolerance (default ±5) as a settlement-lag allowance. So the near-zero relative-tolerance failure mode can't occur.

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Both fixed in the rework: added an all-vacuous exit 2 guard (nothing comparable never reads as PASS), and flipped precedence so divergence(3) > gap(4) with the same rationale as the sibling, so a truncated fetch can't mask a real regression.

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Mooted — the blob-snapshot fetcher and this MetricStatus surface got removed with the 16-cell frame. The new script reads from the subgraphs and mech-analytics API directly.

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Mooted in this PR — the multi-agent aggregation is gone with the reworked gate, so 25 isn't referenced here anymore. But the underlying question is still real: the TRADER_AGENT_IDS 14+25 note lives in etl/agents.py in mech-analytics and is what the rollup will emit for omenstrat, so I'll take confirming 25 with the trader team as a separate follow-up out of this PR.

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -0,0 +1,965 @@
...
+    emit(`    (c) no market_id:        ${counts.no_market_id}`);
+    emit(`        resolved:            ${counts.resolved}`);
+    emit(`        missing key skipped: ${counts.skipped_missing_key}`);
+
+    const check1Truncated = marketplace.truncated || dailyStats.truncated || scoredRows.truncated;
+    let check1Status;
+    if (marketplace.rows.length === 0 && scoredRows.rows.length === 0) {
+      check1Status = 'vacuous';
+      emit('  VACUOUS: no rows on either side in the window.');
+    } else if (open === counts.pending_with_market) {
```
Tanya-atatakai, line 815: a timing question on the core gate: mech-analytics keeps the 24h Omen finality gate (consumer_migration §6 — rows score as resolved only after finalization + 24h), while `profitParticipants` consumes a market as soon as the day's stats record it. so markets resolved in the last ~24h are "consumed" on the subgraph side but still `pending` in analytics — with a 4-day window that's a standing delta on almost every run, and exact equality exits 3 even when the migration is correct. can we bucket the comparison by resolution recency — e.g. also report the subgraph open count with the last 24-48h of settlements NOT consumed, so a red run is mechanically triageable ("delta fully explained by the finality window" vs "real divergence")? wdyt

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -0,0 +1,965 @@
...
+// ---------------------------------------------------------------------------
+// Check 2 — row parity
+// ---------------------------------------------------------------------------
+
+// Multiset diff on (requester, unix timestamp, normalizeTitle(title)).
+// normalizeTitle on both sides matches the consumer's own matching
+// discipline and absorbs title truncation differences between the
+// subgraph's parsedRequest and the lake's question_title.
+const diffRowSets = (subgraphRows, analyticsRows) => {
+  const keyOf = (requester, ts, title) => `${requester}|${ts}|${normalizeTitle(title)}`;
```
Tanya-atatakai: the multiset key uses exact unix-second equality, but live-writer rows carry `requested_at` = mech authorization time, not `blockTimestamp` — they differ by seconds-to-minutes (documented in olas-website `docs/mech-analytics-migration.md`; it's why the consumer dedupes by request id, not timestamp). on backfilled data (`requested_at = block_timestamp`) this passes; the moment live lake writes turn on, every live row becomes a missing+extra pair. and the spec explicitly wants this comparison run on live-writer rows. can we add a second matching pass — same `(requester, normalizeTitle)`, |ts delta| ≤ a few minutes — reported as "matched with ts drift" instead of missing/extra?

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -0,0 +1,965 @@
...
+// (common-util/api/predict/mech-analytics.ts::iterateScoredRows).
+// `since` is >= on requested_at, so windowStart+1s reproduces the
+// subgraph's strict blockTimestamp_gt boundary.
+const fetchScoredRows = async (chainId, windowStartSec, windowEndSec, label) => {
+  const rows = [];
+  let truncated = false;
+  const url = new URL(`${mechAnalyticsBase}/v1/data/scored-rows`);
+  url.searchParams.set('chain_id', String(chainId));
+  url.searchParams.set('since', isoFromUnixSec(windowStartSec + 1));
+  url.searchParams.set('until', isoFromUnixSec(windowEndSec));
```
Tanya-atatakai: tiny, boundary asymmetry: the subgraph side keeps rows with `ts == windowEndSec` (only `ts > windowEndSec` is skipped), but `until` is exclusive on the API — rows in the window's final second exist on one side only:
```suggestion
  url.searchParams.set('until', isoFromUnixSec(windowEndSec + 1));
```

Tanya-atatakai (dismissed):
re-reviewed after the rework — the reframe is exactly right: the three checks now match `consumer_migration.md` §4/§12 Q4, the vacuous guard + exit precedence match the sibling, truncation degrades to gap instead of fake-pass, and the "NOT in this script" scope note about flag-on/off being an ops check is honest. good work turning this around 🙏

approving with one caveat: the 24h-finality timing question inline (check 1 can go chronically red for a reason that isn't a migration bug). as written the gate fails honest and never fake-passes, so the risk is operational noise at cutover, not correctness — but please resolve it before the production run.

**what I verified this round:** staging `/v1/metrics/ai-agent/100/14/instances?days=0` returns a bare array with `agent_address` (check 3 wiring correct); the check-1 consumption logic against `roi-distribution.ts`'s QMR consumption (faithful, incl. per-agent delete and the normalized-index fallback); the exit-code funnel end to end.

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Fixed in ddeb0a55 with a new --finality-lag-hours (default 24) that caps check 1's dailyProfitStatistics date_lte at (now - N hours). That aligns both sides on the resolved universe — settlements still inside the analytics 24h finality gate no longer get consumed on the subgraph side either, so the chronic delta goes away on honest runs. The value is recorded in the metadata header + JSON artifact so the operator can see what buffer was applied.

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Fixed in ddeb0a55. Added id to the marketplace subgraph query and now the multiset key prefers (requester, request_id) when both sides have it — the subgraph Request.id and the lake's request_id are the same identifier (the consumer's own fetcher already dedupes on it). Fallback to (requester, ts, title) only for rows missing an id on either side; a nonzero ts-fallback count on a divergence is flagged in both the console report and the JSON artifact so a post-live-writes ts-drift divergence is distinguishable from a real regression.

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Applied in ddeb0a55 (with a comment tying it to the subgraph's ts == windowEndSec inclusivity). Nice catch.

Tanya-atatakai (approved):

atepem (approved):

PR #553, opened 2026-08-10
fix(metrics): [BDMARK-2246] unfreeze lagging metrics and restore roi-distribution limits

atepem:
## Context

Investigating the greyed-out homepage metrics turned up four separate defects. This PR fixes the code-level ones. Two of the underlying causes are infra and are **not** fixed here — see the bottom.

## What was happening

Production `/api/refresh-metrics/main` and the stored blob showed the homepage values had been frozen for days, and that the values being discarded each hour were *better* than the ones being kept:

| Metric | Shown on site | Freshly fetched | Frozen for |
|---|---|---|---|
| Daily Active Agents | 623 | **644** | 2.2 days |
| OLAS staked | 4,254,952 | **4,250,052** | 2.2 days |
| txns | 19,711,054 | **19,809,613** | 2.2 days |
| Agents deployed | 3,691 | 3,691 | 2.2 days |
| turnover | $107,205.69 | 84,374.08 | 6.4 days |

Cause was `laggingSubgraphs: ["registry:celo"]` / `["staking:celo"]` — Celo behind, not broken.

## Changes

**1. Lag no longer freezes a metric** (`snapshot-storage.ts`, `StaleContent.tsx`)

`mergeWithFallback` discarded the new value on *any* stale status, so one lagging subgraph froze an entire cross-chain aggregate and re-published it hourly, ageing indefinitely. A lagging subgraph still returns real data for its chain, so only a hard failure (`fetchErrors` / `indexingErrors`) or a nil value now triggers fallback. Lag still sets `stale`, so the indicator renders and names the chain.

The tooltip previously read "This metric is outdated" / "Last successful update" — false for a live-but-lagging value — so it now distinguishes frozen from lagging.

Verified across all paths:

```
clean fetch          -> publishes 19,809,613  | no indicator
celo LAGGING only    -> publishes 19,809,613  | indicator YES   <- was frozen at 19,711,054
gnosis FETCH error   -> holds     19,711,054  | indicator YES   <- correctly still frozen
indexing error       -> holds     19,711,054  | indicator YES
null value           -> holds     19,711,054  | indicator YES
```

**2. ROI distribution ran without its function limits** (`vercel.json`)

`e9e3ad9` merged `omenstrat-roi-distribution.ts` + `polystrat-roi-distribution.ts` into `predict-roi-distribution.ts`, updating `crons` but not `functions`. The surviving handler had no entry, so it ran at the default timeout instead of 300s and could not finish its 30-day backfill — the ROI distribution chart rendered empty.

**3. Silent zero-fills** (`main-metrics.ts`, `agent-economies/mech-fees.ts`)

Fetchers read `data.global?.x ?? 0`. A subgraph can answer successfully with `global: null` (e.g. mid-reindex), which is not a GraphQL error — so that chain silently contributed `0` while the metric still published `stale: false`, overwriting the last good blob with a too-low total carrying no indicator. Absent entities are now recorded as fetch errors. An empty `dailyActiveMultisigs` collection is still valid, since no activity is a legitimate answer.

**4. Legacy fee truncation** (`agent-economies/mech-fees.ts`)

`BigInt(x) / BigInt(1e18)` floored away every sub-dollar fraction before the value joined a float total then reported to 6dp. Uses `formatUnits`.

## Expected effect

Next cron after deploy: Daily Active Agents, OLAS staked, txns and Agents deployed resume updating (still amber-flagged while Celo is behind). ROI distribution chart repopulates over subsequent daily runs.

## Not fixed here — needs infra

- **`registry:celo` / `staking:celo` lagging** — needs re-syncing. This PR stops it freezing everything, but the Celo slice stays behind until fixed.
- **`mechFees:gnosis` returns `bad indexers`** — no healthy indexer serving that deployment. Turnover stays correctly frozen until it has coverage, or moves to `*.subgraph.autonolas.tech` (healthy in every probe).

## Trade-off

If a lagging subgraph reindexes from genesis its counter can regress, dragging a published total down where the old behaviour held the last good value. Change 3 catches the common case (missing entity -> fetch error), but a partially-reindexed subgraph returning a small-but-valid count is not caught. A lag-severity threshold would close that gap if it proves to matter.

## Testing

`yarn lint` and `yarn build` pass; `tsc --noEmit` clean. No test runner in this repo. Merge behaviour verified against the real production blob and live subgraph responses as shown above.


common-util/api/agent-economies/mech-fees.ts
```diff
@@ -88,8 +95,9 @@ export const fetchMechFeeMetrics = async () => {
       if (checkSubgraphLag(gnosisBlock, legacy?._meta?.block?.number, 'legacy')) {
         laggingSubgraphs.push('mechFees:legacy');
       }
-      inUsd += Number((BigInt(legacy?.global?.totalFeesIn || '0') / BigInt(1e18)).toString());
-      outUsd += Number((BigInt(legacy?.global?.totalFeesOut || '0') / BigInt(1e18)).toString());
+      // Wei -> xDAI via formatUnits: BigInt division would floor away every sub-dollar fraction.
+      inUsd += Number(formatUnits(BigInt(legacy?.global?.totalFeesIn || '0'), 18));
+      outUsd += Number(formatUnits(BigInt(legacy?.global?.totalFeesOut || '0'), 18));
```
DIvyaNautiyal07: The new-fee loop above (line 73) now catches `global == null` and pushes to `fetchErrors`, but the legacy Gnosis branch here still uses `legacy?.global?.totalFeesIn || '0'` with no equivalent guard. If the legacy subgraph answers with `global: null` (exactly the mid-reindex case the PR is fixing), the code silently adds 0 to `inUsd`/`outUsd`, `createStaleStatus` sees no errors, and `mergeWithFallback` publishes the too-low total with `stale: false`, overwriting the last good blob. Same class of bug the PR is closing, left open on the legacy path.

Fix: wrap lines 99-100 with `if (legacy?.global == null) { console.error('mechFees:legacy: subgraph responded without a global entity'); fetchErrors.push('mechFees:legacy:missingGlobal'); }` before the `inUsd += ...`.

common-util/api/main-metrics.ts
```diff
@@ -402,17 +460,16 @@ export const fetchMechFees = async (): Promise<MetricWithStatus<string | null>>
...
-          const usdValue = Number((data as MechFeesResult).global.totalFeesInUSD || '0');
-          totalFees += usdValue;
-        }
+      const field = isLegacy ? 'totalFeesIn' : 'totalFeesInUSD';
+      const raw = readGlobalField<string>(data.global, field, `mechFees:${source}`, fetchErrors);
+      if (raw === null) return;
+
+      if (isLegacy) {
+        // Legacy mech fees - convert from wei to XDAI
+        totalFees += Number(raw) / 10 ** 18;
```
DIvyaNautiyal07: Divergent wei-to-USD conversion vs `agent-economies/mech-fees.ts:99` for the same legacy Gnosis field. That file now uses `Number(formatUnits(BigInt(x), 18))` (BigInt math, precision-safe). Here it's still `Number(raw) / 10 ** 18` — `Number(wei_string)` casts through a 64-bit float, so once cumulative fees pass ~2^53 wei (~9 ETH-worth) the two functions will report different totals from the same input. Please use `formatUnits` here too so both fee paths agree.

common-util/api/agent-economies/mech-fees.ts
```diff
@@ -68,8 +69,14 @@ export const fetchMechFeeMetrics = async () => {
...
       }
-      inUsd += Number(res.value?.global?.totalFeesInUSD || 0);
-      outUsd += Number(res.value?.global?.totalFeesOutUSD || 0);
+      // `global: null` must not read as zero fees — that drops the chain but reports healthy.
+      if (res.value?.global == null) {
+        console.error(`mechFees:${chain}: subgraph responded without a global entity`);
+        fetchErrors.push(`mechFees:${chain}:missingGlobal`);
+        return;
+      }
+      inUsd += Number(res.value.global.totalFeesInUSD || 0);
```
DIvyaNautiyal07: Nit: `|| 0` re-introduces the same masking one line below the fix — an empty string or numeric `0` field would be silently treated as zero even though the null-guard above just rejected the equivalent case. Consider dropping the `|| 0` (or routing through the `readGlobalField` helper you added in `main-metrics.ts`) so the two situations are handled consistently.

components/ui/StaleIndicator/StaleContent.tsx
```diff
@@ -9,16 +9,23 @@ const formatLocalDate = (timestamp: number | null): string => {
 };
 
 export const StaleMetricContent = ({ status }: StaleIndicatorProps) => {
-  const sources = [
-    ...(status?.indexingErrors || []),
-    ...(status?.fetchErrors || []),
-    ...(status?.laggingSubgraphs || []),
-  ];
+  const hardErrors = [...(status?.indexingErrors || []), ...(status?.fetchErrors || [])];
+  const sources = [...hardErrors, ...(status?.laggingSubgraphs || [])];
```
DIvyaNautiyal07: Nit: `sources` is built once and used once (line 30). Could inline `[...hardErrors, ...(status?.laggingSubgraphs || [])].join(', ')` and drop the intermediate name.

components/ui/StaleIndicator/StaleContent.tsx
```diff
@@ -9,16 +9,23 @@ const formatLocalDate = (timestamp: number | null): string => {
...
+  // Lag alone no longer freezes a metric (see mergeWithFallback), so the value shown is
+  // live and only slightly behind — saying it is outdated would misrepresent it.
+  const isFrozen = hardErrors.length > 0;
```
OjusWiZard: **Blocking — the grey-out still keys off `stale`, so lag-only metrics keep looking dead.**

This correctly splits frozen from lagging for the *tooltip*, but `status.stale` independently drives `text-gray-400` on the value text in ~10 other components:

- `components/HomepageSection/Activity.tsx:105`
- `components/MetricsCard.tsx:111,119`
- `OperatePage/OperateMetrics.tsx`, `OlasTokenPage/TokenHoldersMetric.tsx`, `AgentEconomies/{MechPage,PredictPage,BabydegenEconomyPage,AgentsFunPage}/*`

So in the Celo scenario this PR is fixing: the number updates, but still renders grey — while the tooltip right next to it now says *"Updated: &lt;now&gt; — may be slightly undercounted."* The visual says dead, the text says live, on the same element.

Not a regression (greying happened before too), but it means the motivating symptom — homepage metrics that stop looking stale — isn't actually cleared. `stale` is doing double duty: "show an indicator" and "grey the value out", and only the first is still correct after this change.

Suggest making the distinction an invariant rather than something each consumer re-derives — a `frozen: boolean` on `MetricStatus` computed once in `createStaleStatus`, or a `'fresh' | 'lagging' | 'frozen'` discriminant. Then grey on `frozen`, show the amber icon on `stale`, and this file reads the shared field instead of recomputing it.

common-util/api/agent-economies/mech-fees.ts
```diff
@@ -88,8 +95,9 @@ export const fetchMechFeeMetrics = async () => {
       if (checkSubgraphLag(gnosisBlock, legacy?._meta?.block?.number, 'legacy')) {
         laggingSubgraphs.push('mechFees:legacy');
       }
-      inUsd += Number((BigInt(legacy?.global?.totalFeesIn || '0') / BigInt(1e18)).toString());
-      outUsd += Number((BigInt(legacy?.global?.totalFeesOut || '0') / BigInt(1e18)).toString());
+      // Wei -> xDAI via formatUnits: BigInt division would floor away every sub-dollar fraction.
+      inUsd += Number(formatUnits(BigInt(legacy?.global?.totalFeesIn || '0'), 18));
```
OjusWiZard: **Blocking — the legacy branch still silently zero-fills, 20 lines below the guard that fixes it.**

`legacy?.global?.totalFeesIn || '0'` is exactly the pattern change #3 removes. The new-mech loop above (L73) checks `res.value?.global == null` and pushes a `fetchErrors` entry; this branch doesn't.

If the legacy Gnosis subgraph mid-reindexes and returns `global: null`: fees become `0`, no error recorded, `createStaleStatus` reports `stale: false`, and `mergeWithFallback` sees no hard error and a non-nil value → publishes the too-low total and overwrites the last good blob. That's the precise failure mode the PR describes fixing.

`main-metrics.ts:464` already routes the same legacy field through `readGlobalField`, so the two call sites for this field now disagree with each other.

Same guard as L73 should do it, or reuse `readGlobalField` (see my note on its definition).

common-util/api/agent-economies/mech-fees.ts
OjusWiZard: **Field-level absence is still unguarded.** The new check above catches a missing `global` *entity*, but `totalFeesInUSD || 0` still coerces a present-entity-missing-field response to zero silently — e.g. an older or partially-migrated schema.

`readGlobalField` in `main-metrics.ts` checks per-field for this reason. Worth matching, so the two files agree on what counts as a hard failure.

common-util/api/main-metrics.ts
```diff
@@ -29,6 +29,26 @@ import { getMidnightUtcTimestampDaysAgo } from 'common-util/time';
 const STAKING_CHAINS = Object.keys(STAKING_GRAPH_CLIENTS);
 const REGISTRY_CHAINS = Object.keys(REGISTRY_GRAPH_CLIENTS);
 
+/**
+ * A subgraph can answer successfully with `global: null` (e.g. mid-reindex). Coercing that
+ * to 0 would drop the chain from the sum yet still publish `stale: false`, overwriting the
+ * blob with a too-low number. Record a fetch error so mergeWithFallback keeps the old value.
+ */
+const readGlobalField = <T>(
```
OjusWiZard: Good extraction — but it's only applied to the fetchers in this file. The same `global: null → 0` pattern is still live on the same `saveSnapshot` → `mergeWithFallback` publish path elsewhere:

- `common-util/api/agent-economies/mech.ts:134-135` — `Number(data.global?.totalRequests ?? 0)` / `totalDeliveries ?? 0`. Highest impact of the four: same category as the file you did fix, and it feeds the derived `otherTxsValue = totalRequests - known` bucket, so one nulled chain distorts two published numbers.
- `common-util/api/other-metrics/build.ts:20` — `totalBuilders`
- `common-util/api/other-metrics/govern.ts:41` — `token?.balance ?? '0'`
- `common-util/api/other-metrics/token-holders.ts:33` — `holderCount ?? 0`

All pre-existing, so a fair scope call either way — but the PR describes change #3 as fixing "silent zero-fills" generally, and these are the same bug on the same path. Either pull them in, or narrow the claim in the description so they don't read as covered.

If you do: this helper has no `main-metrics` dependency and would sit naturally in `common-util/graphql/metric-utils.ts` next to `createStaleStatus`.

common-util/api/main-metrics.ts
```diff
@@ -29,6 +29,26 @@ import { getMidnightUtcTimestampDaysAgo } from 'common-util/time';
...
+  source: string,
+  fetchErrors: string[]
+): T | null => {
+  const value = globalEntity?.[field];
+  if (value === undefined || value === null) {
+    console.error(`${source}: subgraph responded without global.${field}`);
+    fetchErrors.push(`${source}:missingGlobal`);
+    return null;
+  }
+  return value as T;
```
OjusWiZard: `value as T` is an unchecked cast with nothing connecting the stringly-typed `field` to the caller-chosen `T`.

I checked all 5 call sites (L164, L235, L306, L376, L464) — every one currently matches the field's real type, so **no live bug**. Flagging it as a footgun rather than a defect.

The way it bites later: `operatorsByChains` is `number[]` and gets summed with `sum + operatorsByChain`. A future call-site copy-paste passing `<number>` for a field that's actually a numeric string type-checks fine and silently does string concatenation on a headline metric — no compiler signal, no throw, just a wrong number published. `strict: false` in tsconfig means there's no second line of defence.

Constraining the field to `keyof` so `T` is inferred rather than asserted would close it:

```ts
const readGlobalField = <TGlobal, K extends keyof TGlobal>(
  globalEntity: TGlobal | null | undefined,
  field: K,
  source: string,
  fetchErrors: string[]
): TGlobal[K] | null => { ... }
```

The `*Result` types already declared in this file have the correct field types, so call sites would get real checking instead of documentation.

common-util/snapshot-storage.ts
```diff
@@ -57,7 +57,14 @@ const mergeWithFallback = (newData: unknown, oldData: unknown, path: string = ''
     const newMetric = newData as MetricWithStatus<unknown>;
     const oldMetric = isMetricWithStatus(oldData) ? (oldData as MetricWithStatus<unknown>) : null;
 
-    const newValueIsInvalid = isNil(newMetric.value) || newMetric.status?.stale;
+    // Only a hard failure makes the new value untrustworthy. A lagging subgraph still
+    // returns real data for its chain — just a little behind — so the fresh aggregate is
+    // closer to the truth than a frozen one that ages indefinitely. Falling back on lag
+    // alone was making published metrics *less* accurate the longer a chain stayed behind.
+    const hasHardError =
```
OjusWiZard: The change itself is right — this is the core fix and I agree with the reasoning in the comment.

One thing: this predicate now exists twice, both added in this PR. Here as a boolean OR, and in `StaleContent.tsx:12,16` as an array concat + `.length > 0`. Same rule, two shapes. If a third status field is ever added, one copy gets updated and the other quietly doesn't — and these two specifically decide whether a metric is frozen and whether the user is told it's frozen, so a drift between them is exactly the kind of inconsistency that's hard to notice.

Worth exporting one predicate from `metric-utils.ts`:

```ts
export const hasHardError = (status?: Pick<MetricStatus, 'fetchErrors' | 'indexingErrors'> | null) =>
  (status?.fetchErrors?.length ?? 0) > 0 || (status?.indexingErrors?.length ?? 0) > 0;
```

That's also the natural place to hang the `frozen` flag the UI needs (see my `StaleContent.tsx` comment) — one definition feeding the merge, the tooltip, and the grey-out.

To be clear, `stale` in `createStaleStatus` should stay broader (it includes lag) — that's a genuinely different question ("show an indicator at all"), not a third copy of this one.

CLAUDE.md
```diff
@@ -66,7 +66,7 @@ Coverage varies by category (e.g. `BABYDEGEN_GRAPH_CLIENTS` only has Optimism +
 
 Categories: `main`, `predict`, `agent-economies`, `other`. Plus daily snapshots for `predict-roi-distribution` (per agent) and `predict-tool-accuracy`.
 
-**`MetricWithStatus<T>` shape**: every metric is `{ value, status: { stale, lastValidAt, indexingErrors, fetchErrors, laggingSubgraphs } }`. Use `createStaleStatus`, `getFetchErrorAndCreateStaleStatus`, and `checkSubgraphLag` from `common-util/graphql/metric-utils.ts` to populate it. `CHAIN_CONFIG` in `common-util/constants.ts` defines the per-chain `lagLimit` (calibrated for ~12h indexing tolerance).
+**`MetricWithStatus<T>` shape**: every metric is `{ value, status: { stale, lastValidAt, indexingErrors, fetchErrors, laggingSubgraphs } }`. Use `createStaleStatus`, `getFetchErrorAndCreateStaleStatus`, and `checkSubgraphLag` from `common-util/graphql/metric-utils.ts` to populate it. `CHAIN_CONFIG` in `common-util/constants.ts` derives each chain's `lagLimit` from a measured `blockTimeSec` and a `lagToleranceHours` that scales with the chain's share of the totals (12h for ethereum/gnosis, 48h for the sub-2% chains). Set the tolerance in hours — never hand-write a block count.
```
OjusWiZard: This line is updated correctly, but **line 64 in the same file still describes the pre-PR behaviour**:

> `mergeWithFallback` (keeps the last valid value when the new value is null/**stale** and marks `status.stale = true`)

After this PR the trigger is null-or-hard-error; lag alone no longer falls back. That's the exact mechanism this PR changes, so it's the one line most worth keeping accurate — anyone (or any agent) reading CLAUDE.md for the merge semantics gets the old model.

OjusWiZard (commented):
Reviewed the four stated changes against the surrounding code. The diagnosis is right and the `mergeWithFallback` fix is the correct call — publishing a live-but-lagging aggregate really is better than re-publishing a frozen one that ages indefinitely. `vercel.json` is now fully consistent (7 `functions` ↔ 7 files ↔ all cron paths), and the `formatUnits` fix addresses a genuine truncation (`BigInt(x) / BigInt(1e18)` floored `123456.789` → `123456`).

Two things I'd want resolved before merge, plus a few smaller ones — left inline.

The headline concern: the fix stops one layer short of the symptom that motivated it. `status.stale` still independently drives the grey-out on the value text, so lag-only metrics will keep rendering greyed even though they're now fresh — with the new tooltip on the same element saying "Updated: &lt;now&gt;". Details on `StaleContent.tsx`.

---

**On the testing note** — "No test runner in this repo" isn't quite right. `scripts/*.test.mjs` run under `node --test` via the `parity-script-tests` CI job, and that job is in `all-checks-passed.needs`, so it's a required, merge-blocking check. Its own comment says it exists because *"two prior wiring regressions shipped because these helpers had no CI coverage."*

`mergeWithFallback` and `readGlobalField` are pure, synchronous, and gate every number on the public dashboard — and neither is exported. The verification table in the PR description is already a table-driven test matrix, just executed by hand:

| case | value | fetchErrors | lagging | old blob | expect value | expect `stale` |
|---|---|---|---|---|---|---|
| clean | present | [] | [] | any | new | `false` |
| lag only | present | [] | `['celo']` | any | **new** | `true` |
| fetch error | present | `['x']` | any | valid | **old** | `true` |
| indexing error | present | [] | any | valid | **old** | `true` |
| nil value | `null` | [] | [] | valid | **old** | `true` |

Porting that costs no new dependencies and follows the pattern the team already established for exactly this failure mode. A `readGlobalField` case for `{global:{field:0}}` would also lock in that a legitimate zero isn't mistaken for a missing entity.

**One more, not anchorable to the diff:** `getSnapshot` swallows every failure and returns `null`, so `saveSnapshot` skips `mergeWithFallback` entirely and writes the raw fetch straight to the blob. A transient blob-read failure coinciding with a real subgraph error publishes the degraded value anyway, and the route still returns `200 {success:true}`. Pre-existing, but it's the escape hatch around the protection this PR adds — worth at least distinguishing "no snapshot yet" from "load failed".

**Also worth noting:** the description covers 4 changes, but the branch carries two more commits — `9fc8252` (lag tolerance rescaled across all 8 chains; arbitrum goes 11.1h → 48h, a 4.3× loosening) and `b0ec6ae` (nanoid pin). Reviewers working from the description would miss both. Arbitrum is the one chain in the 48h bucket whose traffic share isn't cited in the new comment block.

components/ui/StaleIndicator/StaleContent.tsx
atepem: Fixed in cc26950 — this was the right catch, the PR stopped one layer short of its own motivating symptom.

Took the discriminant option: `MetricStatus` now carries `frozen` (the published value is a held-over fallback), computed once in `createStaleStatus`/`resolveMergedMetric`. `stale` stays broader and keeps driving the amber icon; `frozen` drives the grey-out, switched across all 10 consumers you listed. `MetricsCard` and `OperateMetrics` kept a separate `isStale` for the indicator alongside a new `isValueFrozen` for the text.

Made it read-through rather than raw so old blobs still work: `isFrozen(status)` returns `status.frozen ?? hasHardError(status)`, deriving from the error arrays when the field is absent. That means no `SCHEMA_VERSIONS` bump and no window where pre-existing snapshots render wrong.

common-util/api/agent-economies/mech-fees.ts
atepem: Fixed in cc26950. Both legacy fields now route through `readGlobalField`, same as the per-chain loop. You were right that the two call sites for the same field disagreed.

common-util/api/agent-economies/mech-fees.ts
atepem: Fixed in cc26950 — dropped the `|| 0` and moved both fields onto `readGlobalField`, so a present-entity-missing-field is treated the same as a null entity.

common-util/api/main-metrics.ts
atepem: Pulled them in rather than narrowing the claim (cc26950), and moved the helper to `metric-utils.ts` as suggested.

- `mech.ts` — both fields via `readGlobalField`; agreed this was the highest impact given the derived `other` bucket.
- `build.ts` / `govern.ts` — these go through `executeGraphQLQuery`, whose `transform` has no `fetchErrors` array to push to. Returning `null` instead of `0`/`"0"` gets the same outcome without changing that signature: nil value → fallback → `frozen: true`.
- `token-holders.ts` — has its own `{ count, error }` shape, so it records `tokenHolders:<key>:missingGlobal`.

common-util/api/main-metrics.ts
atepem: Applied in cc26950 — `<TGlobal, K extends keyof TGlobal>` returning `TGlobal[K] | null`, and dropped the explicit type args at all 5 call sites so they infer from the declared `*Result` types.

One knock-on worth flagging: the `mechFees` path picked its field dynamically (`isLegacy ? "totalFeesIn" : "totalFeesInUSD"`) against a union of two global types, which does not infer under `keyof`. Split it into explicit branches — which also gave the natural place to fix the `formatUnits` divergence from your other comment.

common-util/snapshot-storage.ts
atepem: Done in cc26950 — exported `hasHardError` from `metric-utils.ts` with your signature; the merge and the tooltip both call it now.

Took the follow-on too: `frozen` hangs off the same predicate, so one definition feeds the merge, the tooltip and the grey-out. And agreed on `stale` staying broader — left it including lag, since "show an indicator at all" is genuinely a different question.

CLAUDE.md
atepem: Fixed in cc26950 — line 64 now describes the null-or-hard-error trigger and says explicitly that lag sets `stale` but not `frozen`. Also added a line to the `MetricWithStatus` section telling consumers to grey on `isFrozen(status)`, never `status.stale`, since that is the trap this PR just walked into.

common-util/api/agent-economies/mech-fees.ts
atepem: Fixed in cc26950 — same guard as the new-fee loop, via `readGlobalField`.

common-util/api/main-metrics.ts
atepem: Fixed in cc26950 — both paths use `formatUnits` now. Agreed on the reasoning: `Number(weiString)` casting through a float meant the two functions could report different totals from identical input.

common-util/api/agent-economies/mech-fees.ts
atepem: Fixed in cc26950 — dropped the `|| 0` and routed through `readGlobalField`, so the two cases are handled consistently.

components/ui/StaleIndicator/StaleContent.tsx
atepem: Left as-is deliberately. `sources` is now built from `hardErrors`, which is also used on its own for the frozen/lagging branch — inlining would mean spreading `hardErrors` twice. Happy to change it if you would still prefer the inline.

components/ui/StaleIndicator/StaleContent.tsx
DIvyaNautiyal07: **P1** — this local `isFrozen` was not migrated to the new `isFrozen()` helper in `metric-utils.ts`. It computes `hardErrors.length > 0`; the helper computes `status?.frozen ?? hasHardError(status)`. They diverge exactly in the case this PR just introduced in `build.ts` / `govern.ts`: transform returns `null` → `resolveMergedMetric` sets `frozen: true, stale: true` on the returned status while `fetchErrors` stays `[]`. Consumers (`MetricsCard`, `OperateMetrics`, etc.) then correctly grey the number via `isFrozen(status)`, but this tooltip's local `isFrozen` evaluates to `false` and renders *"Some sources are behind the chain, so this may be slightly undercounted. Updated: ..."* — promising live data underneath a greyed number that's actually held-over.

Fix: `import { isFrozen } from 'common-util/graphql/metric-utils'` and rename the local variable (or drop it in favour of the helper).

common-util/graphql/metric-status.ts
```diff
@@ -0,0 +1,95 @@
...
+ * A subgraph can answer successfully with `global: null` (e.g. mid-reindex). Coercing that
+ * to 0 would drop the chain from a cross-chain sum yet still publish `stale: false`,
+ * overwriting the blob with a too-low number. Record a fetch error instead, so the metric
+ * freezes and the merge keeps the last good value. `K extends keyof TGlobal` makes the
+ * return type follow the field rather than a caller-supplied assertion.
+ */
+export const readGlobalField = <TGlobal, K extends keyof TGlobal>(
+  globalEntity: TGlobal | null | undefined,
+  field: K,
+  source: string,
```
DIvyaNautiyal07, line 42: **P2** — the push string doesn't carry the field name, so callers that read multiple fields per source will emit identical duplicates on a null global. `mech-fees.ts:75-76`, `mech-fees.ts:101-107` and `mech.ts:138-139` all read two fields against the same source; when `global` is null both `readGlobalField` calls push `${source}:missingGlobal`, and `StaleContent` renders `Affected Sources: mechFees:gnosis:missingGlobal, mechFees:gnosis:missingGlobal`. Also collapses the diagnostic: a single-field miss looks identical to a null-entity miss.

Fix: push `${source}:missing:${String(field)}` — dedups naturally by field name and preserves which field was absent.

common-util/api/other-metrics/build.ts
```diff
@@ -17,7 +17,10 @@ const fetchTotalBuilders = async (): Promise<MetricWithStatus<number | null>> =>
     chain: 'ethereum',
     query: totalBuildersQuery,
     source: 'build:totalBuilders',
-    transform: (data) => Number(data.global?.totalBuilders || 0),
+    // null (not 0) on an absent entity, so mergeWithFallback holds the last good value
+    // instead of publishing a zero as healthy.
+    transform: (data) =>
+      data.global?.totalBuilders == null ? null : Number(data.global.totalBuilders),
```
DIvyaNautiyal07: Nit — returning `null` from `transform` correctly triggers `mergeWithFallback` to hold the old value with `frozen: true`, but no `fetchError` is pushed anywhere. The metric ends up greyed with an empty "Affected Sources" line (nothing in `fetchErrors`, `indexingErrors`, or `laggingSubgraphs`), so the user sees a held-over value with no explanation. Same pattern in `govern.ts:41`. Contrast with the multi-chain fetchers that use `readGlobalField`, which do push. Consider either letting `executeGraphQLQuery` record a fetchError when `transform` returns null, or threading a `fetchErrors` array through so the missing entity is named in the tooltip.

components/ui/StaleIndicator/StaleContent.tsx
OjusWiZard: The `frozen` split got wired into all ~10 grey-out sites but not into this one — the tooltip still derives its own predicate and never reads `status.frozen`.

Local: `hardErrors.length > 0` (i.e. `hasHardError`). Shared: `isFrozen(status)` = `status.frozen ?? hasHardError(status)`. They disagree whenever `frozen: true` with empty error arrays.

That state is reachable through code added in this same commit — `build.ts:23` and `govern.ts:42` now return `value: null` on an absent entity *without* recording an error, and `resolveMergedMetric` freezes on `isNil(value)` independently of `hasHardError`. Running the real module on it:

```
published value : 42 (held-over)   frozen: true   stale: true
grey-out  isFrozen(status)      -> true
tooltip   hardErrors.length > 0 -> false
tooltip   affected sources      -> []
```

So `totalBuilders` / `lockedOlas` render greyed while this tooltip says "may be slightly undercounted" and "Updated: <a stale timestamp>", listing no sources — the two halves of the same card contradicting each other.

There's a structural reason it was easy to miss: **`components/ui/StaleIndicator/types.ts` declares its own local `MetricStatus` with no `frozen` field**, so this component can't read it without fixing that type first. Worth checking `PlatformActivitySection.tsx:21` too — it derives its domain types from that same local duplicate (`type MetricStatus = StaleIndicatorProps['status']`), so five status fields there are statically missing `frozen`, hidden only because the field is optional.

Fix is both files: import the canonical `MetricStatus` from `common-util/graphql/types.ts`, then use the shared `isFrozen`.

common-util/graphql/metric-status.ts
```diff
@@ -0,0 +1,95 @@
...
+ * Re-exported from metric-utils.ts; import from there in app code.
+ */
+import type { MetricStatus, MetricWithStatus } from './types';
+
+const isNil = (v: unknown): boolean => v === null || v === undefined;
+
+/**
+ * A source failed outright, as opposed to merely running behind. Only this forces a
+ * fallback to the previous value — lag still publishes live data. Single definition so the
+ * merge, the tooltip and the grey-out cannot drift apart.
```
OjusWiZard, line 16: This claim isn't true yet — the tooltip doesn't call `hasHardError`. `StaleContent.tsx:12,16` still recomputes it inline (see my comment there).

Worth fixing the code rather than the comment, since the comment describes the right design. But please don't leave the assertion standing on its own — a future reader will trust it and skip the check.

common-util/graphql/types.ts
```diff
@@ -10,7 +10,18 @@ export type WithMeta<T> = T & {
...
+   * Broader than `frozen`: a lagging subgraph is stale but still publishes live data.
+   */
   stale: boolean;
+  /**
+   * The published value is a held-over fallback rather than this run's data, because a
+   * source hard-failed or returned nothing. Drives the greyed-out value text. Optional so
+   * snapshots written before this field existed still deserialize — read it through
+   * `isFrozen()` in metric-utils, which derives it from the error arrays when absent.
+   */
+  frozen?: boolean;
```
OjusWiZard: Now that `createStaleStatus` **and** `resolveMergedMetric` both set `frozen` unconditionally, every in-repo producer writes it — so the `?? hasHardError(status)` fallback in `isFrozen` is effectively dead for anything written after this commit.

The stated justification (old snapshot blobs) doesn't quite hold either: `getSnapshot` only shallow-checks `'data' in x && 'timestamp' in x` and then casts, so deserialised blobs are never type-checked against `MetricStatus` anyway. Optionality here isn't buying back-compat safety — TypeScript was never guarding that boundary.

Concrete argument for making it required: it would have turned the `StaleIndicator/types.ts` drift into a compile error instead of letting it pass silently via structural subtyping. The optional field is what let that duplicate type survive.

common-util/api/agent-economies/mech.ts
```diff
@@ -131,8 +132,15 @@ const fetchMechGlobals = async (): Promise<
           laggingSubgraphs.push(`marketplace:${chain}`);
         }
 
-        totalRequests += Number(data.global?.totalRequests ?? 0);
-        totalDeliveries += Number(data.global?.totalDeliveries ?? 0);
+        // Zero-filling here would also distort the derived `other` bucket
+        // (totalRequests - known), so one nulled chain would skew two published numbers.
+        const source = `marketplace:${chain}`;
+        const requests = readGlobalField(data.global, 'totalRequests', source, fetchErrors);
```
OjusWiZard, line 138: This guard is right. But the derived `otherTxs` bucket further down in the same file still leaks the failure it's meant to catch — flagging it here since there's no diff line to anchor to.

`fetchMechMetrics` (~L312):

```ts
const agentsfunTxsValue = agentsfunTxs.value ?? 0;   // hard-failed -> silently 0
const known = predictTxs + contributeTxs + governatooorrTxs + agentsfunTxsValue;
const otherTxsValue = Math.max(0, totalRequestsValue - known);
```

and `otherTxs.status` (~L340-356) is rebuilt from **only** `globals.status` and `categorized.status` — `agentsfunTxs.status` is never merged in.

So if only the agents.fun subgraph fails: its value nulls to `0`, `known` comes out too small, `otherTxs` is inflated, and it publishes with `stale: false, frozen: false` — straight past `hasHardError` and over the last good blob. The standalone `agentsfunTxs` metric freezes correctly; only the derived bucket leaks.

Pre-existing and not something this PR touched, so no objection to it landing as-is — but it's the same failure class this PR is closing, and it's the one instance still live. Probably its own PR: fold `agentsfunTxs.status` into the `createStaleStatus` call, and gate `otherTxs.value` on `agentsfunTxs.value != null` the way it already gates on `globals.value && categorized.value`.

common-util/api/other-metrics/build.ts
OjusWiZard: Returning `null` instead of `0` is the right call given `transform` has no `fetchErrors` array to push to — agreed with your reasoning on the earlier thread.

One gap: unlike `readGlobalField`, this path doesn't `console.error`. So when the entity does go absent there's no server-side trace at all — and if this write happens to land on the merge-bypass path (no prior snapshot / blob read failure), the page just shows `--` with no indicator, no tooltip, and nothing in the logs. Same applies to `govern.ts:42`.

A one-line `console.error` before the `null` would keep it diagnosable.

common-util/snapshot-storage.ts
```diff
@@ -4,6 +4,7 @@ import { ExplorerMetricsData } from 'common-util/api/explorer';
 import { MainMetricsData } from 'common-util/api/main-metrics';
 import { OtherMetricsData } from 'common-util/api/other-metrics';
 import { PredictMetricsData } from 'common-util/api/predict';
+import { resolveMergedMetric } from 'common-util/graphql/metric-utils';
 import { isMetricWithStatus, MetricWithStatus } from 'common-util/graphql/types';
 import { isNil, isPlainObject } from 'lodash';
```
OjusWiZard: `isNil` is unused now that the nil check moved into `resolveMergedMetric` — `isPlainObject` is still used, `isNil` isn't referenced anywhere else in the file. Lint doesn't flag it, so CI won't catch it.

common-util/api/main-metrics.ts
```diff
@@ -402,17 +437,28 @@ export const fetchMechFees = async (): Promise<MetricWithStatus<string | null>>
...
-        } else {
-          // New mech fees (gnosis, base) - already in USD
-          const usdValue = Number((data as MechFeesResult).global.totalFeesInUSD || '0');
-          totalFees += usdValue;
-        }
+      if (isLegacy) {
+        // Legacy mech fees — wei-denominated. Converted with formatUnits (BigInt math) to
+        // match agent-economies/mech-fees.ts: `Number(weiString) / 1e18` casts through a
+        // 64-bit float first, so the two would disagree on the same input past 2^53 wei.
+        const raw = readGlobalField(
```
OjusWiZard: Minor: the split duplicates the `readGlobalField` call and the `if (raw === null) return;` guard across both branches, when the only thing that actually differs is the final conversion. Since `keyof` inference was the reason for splitting, you can keep one call and branch just on the arithmetic:

```ts
const raw = isLegacy
  ? readGlobalField(data.global, 'totalFeesIn', source, fetchErrors)
  : readGlobalField(data.global, 'totalFeesInUSD', source, fetchErrors);
if (raw === null) return;
totalFees += isLegacy ? Number(formatUnits(BigInt(raw), 18)) : Number(raw);
```

Entirely optional — current version is correct, just a few lines longer.

OjusWiZard (dismissed):
Approving. Every item from the last round is genuinely addressed, and the shape you landed on is better than what I suggested — extracting the pure decisions into `metric-status.ts` with no runtime imports means they're testable under `node --test` with zero new dependencies, and splitting `frozen` from `stale` fixes the root confusion rather than patching the symptom.

Verified rather than assumed:

- `node --test common-util/graphql/metric-status.test.mjs` → 17/17, on the pinned Node 22.22.3 from `.nvmrc` (type-stripping works unflagged there, so no CI break risk)
- `tsc --noEmit` clean — the generic `readGlobalField<TGlobal, K extends keyof TGlobal>` and all five rewritten call sites type-check
- `metric-status:test` runs inside `parity-script-tests`, which is in `all-checks-passed.needs` — genuinely blocking
- Traced `resolveMergedMetric` on first-run/no-snapshot, legacy-blob-without-`frozen`, lag-only, hard-error and nil-value paths; all correct
- `token-holders.ts` returning `count: 0` is harmless — it always pushes a `fetchErrors` entry too, so the partial sum is discarded at merge and never surfaces
- The mech-fees legacy/new split introduces no double-counting

The comments below are all non-blocking. Two are worth doing before merge (the tooltip, and the doc comment that currently asserts it's already done); the rest are follow-ups.

---

**Not anchorable to the diff, for the record:**

**Merge bypass on a missing prior snapshot.** `getSnapshot` returns `null` on a first run, a `SCHEMA_VERSIONS` bump, *or* a transient blob-read failure, and `saveSnapshot` then skips `mergeWithFallback` entirely and writes the raw fetch. On that path every hard-errored metric in the category loses its hold-last-good-value fallback, for a reason unrelated to the metric itself. Distinguishing "no snapshot yet" from "load failed" would close it. Pre-existing; raised last round; still open.

**`mergeWithFallback`'s recursive walk is still unexported and untested.** You extracted the leaf decision, which was the valuable half — but the tree-walk around it (object recursion, the `allKeys` old-key carry-forward, and the `Array.isArray` short-circuit that returns arrays as-is with no fallback applied) has no coverage. I checked every snapshot shape and none currently nests a `MetricWithStatus` inside an array, so the array branch isn't reachable today — it's a trap for whoever first adds `MetricWithStatus[]`, not a live bug.

**`constants.ts`** still asserts traffic shares (~96.6% gnosis, optimism 1.69%, …) and "measured" block times with no source, date, or refresh path. They'll silently rot. Either cite where they came from and when, or soften the wording to "configured".

**Follow-up worth its own PR:** `fetchTotalOlasStaked` / `fetchTransactions` / `fetchTotalOperators` / `fetchAtaTransactions` are now structurally identical — allSettled → lag check → `readGlobalField` → push → reduce. With the helpers shared and tested, collapsing them into one generic multi-chain aggregator became a much smaller step than it was before this PR.

components/ui/StaleIndicator/StaleContent.tsx
atepem: Fixed in 5cf1809 — correct on both the bug and the structural reason it was missable.

`StaleContent` now calls the shared `isFrozen`, and `StaleIndicator/types.ts` re-exports the canonical `MetricStatus` instead of redeclaring it.

Your point about the duplicate type surviving because the field was optional turned out to be sharper than it looked: making `frozen` required surfaced **two more** redeclarations you had not listed — `HomepageSection/Activity.tsx:30` and an inline `status?: {...}` in `MetricsCard.tsx:71`. Both were invisible via structural subtyping. All three now use the canonical type, which fixes `PlatformActivitySection` transitively since it derives from `StaleIndicatorProps`.

common-util/graphql/metric-status.ts
atepem: Fixed the code rather than the comment (5cf1809) — the tooltip calls `isFrozen`, which calls `hasHardError`, so the claim now holds. Agreed on not leaving the assertion standing; that is exactly how the next reader gets misled.

common-util/graphql/types.ts
atepem: Made it required in 5cf1809. Your argument won on the evidence: it immediately turned up two more local redeclarations beyond the one you flagged (`HomepageSection/Activity.tsx`, `MetricsCard.tsx`), both previously passing via structural subtyping.

Agreed the back-compat justification did not hold — `getSnapshot` shallow-checks and casts, so TypeScript was never guarding that boundary. I kept the `?? hasHardError(status)` fallback in `isFrozen` anyway, but on the honest grounds: it is runtime defence for blobs written before the field existed, not type-level back-compat. Comment updated to say that.

common-util/api/other-metrics/build.ts
atepem: Added `console.error` on both null-transform paths in 5cf1809. The compounding case you describe — merge-bypass plus no log — was the convincing part: `--` on the page with nothing anywhere to explain it.

common-util/snapshot-storage.ts
atepem: Removed in 5cf1809. And noted that lint would not have caught it.

common-util/api/main-metrics.ts
atepem: Applied in 5cf1809 — one `readGlobalField` call, one guard, branch only on the conversion. Reads better than what I had.

components/ui/StaleIndicator/StaleContent.tsx
atepem: Fixed in 5cf1809 — imports the shared `isFrozen` and the local variable is renamed to `frozen`. Your reproduction was exactly right: the divergence is reachable through `build.ts`/`govern.ts` returning null with empty `fetchErrors`. Added a regression test pinning `frozen: true` with `hasHardError === false` and `isFrozen === true`.

common-util/graphql/metric-status.ts
atepem: Fixed in 5cf1809 — now pushes `${source}:missing:${String(field)}`, with a test asserting two absent fields on one source stay distinct rather than rendering as identical duplicates.

common-util/api/other-metrics/build.ts
atepem: Added the `console.error` to both (5cf1809).

Did not thread a `fetchErrors` array through `executeGraphQLQuery` for now — it would change that signature for all callers, and the value is still correctly held with `frozen: true`. The cost is the empty "Affected Sources" line you describe, which is a real gap; I would rather do it as its own change than widen this PR further. Happy to file it.

common-util/api/agent-economies/mech.ts
atepem: Agreed, and thanks for tracing it — `otherTxs` rebuilding its status from only `globals` and `categorized` while `agentsfunTxs` nulls to `0` is the same failure class this PR closes, and it is genuinely still live.

Leaving it out deliberately: it is pre-existing, needs both a status merge and a value gate, and this PR is already large. Filing it as a follow-up with your analysis rather than folding it in.

components/ui/StaleIndicator/StaleContent.tsx
atepem: Correcting my earlier reply on this thread — the justification I gave no longer holds. `hardErrors` is gone as of 5cf1809 (the frozen check moved to the shared `isFrozen`), so `sources` really is built once and used once now, exactly as you said.

Still leaving it, but on plainer grounds: inlining a three-way spread into the JSX reads worse than the named binding. Say the word and I will change it.

common-util/api/other-metrics/build.ts
atepem: Done properly in 3025c39 rather than deferred — `executeGraphQLQuery` now passes a `fetchErrors` sink into `transform`, so `build.ts` and `govern.ts` route through `readGlobalField` like every other call site. The absent entity is named in the tooltip instead of producing an empty "Affected Sources" line.

Changing the signature turned out cheap: `transform` is optional-arity, so the other call sites needed no edit.

atepem: Superseded by 3025c39 — rather than just logging, `transform` now receives a `fetchErrors` array and both paths go through `readGlobalField`. That gives the `console.error` you asked for *and* names the source in the tooltip, so the merge-bypass scenario you described no longer ends with nothing anywhere.

common-util/api/agent-economies/mech.ts
atepem: Fixed in 3025c39 rather than split out — your trace was exact. `agentsfunTxs` now gates the value alongside `globals` and `categorized`, and its status is merged into the `createStaleStatus` call, so a failed agents.fun subgraph freezes the derived bucket instead of inflating it and publishing as healthy.

OjusWiZard (approved):
Re-approving `3025c39` — the previous approval was dismissed by the two follow-up commits. Everything raised across all three rounds is now addressed, and several of the fixes are better than what was suggested.

Re-verified on this head rather than assumed:

- `node --test common-util/graphql/metric-status.test.mjs` → 25/25 (up from 17), now covering the tree walk as well as the leaf decision
- `tsc --noEmit` clean; full CI green including the blocking `parity-script-tests` job
- The new `saveSnapshot` throw is handled everywhere: all seven `refresh-metrics` handlers catch it and return 500, so a failed merge shows up as a cron failure with the existing blob untouched. Every `overwrite: true` caller bypasses `loadSnapshot` and can't reach it.
- The `mergeSnapshotTree` move is behaviour-preserving vs the old `mergeWithFallback`, and the new array warning can't fire on any current snapshot shape (checked each one — arrays only ever live inside a metric's `value`)

Particularly worth calling out: the `found`/`absent`/`error` tri-state is the right shape for the merge-bypass problem — refusing the write and letting the next cron retry is strictly better than either the old silent overwrite or a blanket failure. And extracting `mergeSnapshotTree` into the import-free module so the recursive walk became testable was the more useful half of that change.

---

Nothing below blocks the merge; recording it so it isn't lost.

**Worth its own PR — the read path still has the bug the write path just fixed.** `getSnapshot` still folds `absent` and `error` into a single `null`, and all 13 `getStaticProps` bake that into props:

```ts
// pages/index.tsx:36-47
const metricsSnapshot = await getSnapshot({ category: 'main' });
const metrics = metricsSnapshot?.data ?? null;
return { props: { metrics }, revalidate: REVALIDATE_DURATION };
```

A transient blob read during ISR revalidation returns successfully with `metrics: null`, so Next.js replaces the good cached page with an empty one. Throwing would make it keep serving the last good page and retry. Entirely pre-existing — `pages/` has no changes in this PR — but the tri-state you just built makes it a small fix, and the argument for it is already written in this PR's own comments.

**`babydegen.ts` didn't pick up the new `fetchErrors` sink.** `transform(data, fetchErrors)` exists so a null-returning transform can name its source in the tooltip; it's used at 2 of the 9 transform call sites. All six in `babydegen.ts` still return `null` without recording anything, so those metrics freeze showing "This metric is outdated due to some issues" with an empty Affected Sources line — the symptom the sink was added to remove.

**Two test gaps on new logic.** `saveSnapshot`'s tri-state now decides whether an entire category publishes and has no coverage — but it's extractable exactly the way `mergeSnapshotTree` was (`decideSaveAction(result, overwrite)` touches no blob I/O). Nothing tests that errors a transform pushes actually reach the returned status either.

**Smaller things:**
- A metric whose shape changes to a non-metric between runs merges into a hybrid: `new {count:7} + old {value:42,status:{…}}` → `{count:7, value:42, status:{…}}`, which is metric-shaped again on the next run. Needs a shape change without a `SCHEMA_VERSIONS` bump, and that map is still `{}` — the guard is entirely manual.
- `{a: undefined}` drops the old value where `{}` carries it forward. Latent; nothing emits explicit `undefined` today.
- `isMetricWithStatus` in `types.ts` is now dead — `metric-status.ts` reimplements it privately as `isMetricLike` rather than importing it (which wouldn't break the no-runtime-imports rule, since `types.ts` is dependency-free).
- The array test asserts only that `console.warn` fired, not the pass-through it's named for — it'd still pass if that behaviour changed.
- `` `list failed: ${(error as Error)?.message}` `` can render `"list failed: undefined"`; the sibling branch has a `?? 'read failed'` fallback.
- `frozen` being required is enforced by convention — a direct `status.frozen` read would compile and misbehave on an old blob. Nothing does today.

ROMAN-VALORY (approved):

DIvyaNautiyal07 (approved):
LGTM on the delta. All prior findings addressed with tests to guard against regression:

- `StaleContent` now uses the shared `isFrozen` helper; new test covers the empty-error-arrays case.
- `readGlobalField` pushes field-qualified strings; duplicate-source case tested.
- `executeGraphQLQuery.transform` receives a `fetchErrors` sink; `build.ts` / `govern.ts` name their source in the tooltip.
- `otherTxs` derived bucket now gates on `agentsfunTxs.value` and merges its status.
- `MetricStatus` canonicalized; `frozen` made required so future producers can't silently drop it.
- `snapshot-storage` refactor: `loadSnapshot` distinguishes `absent` from `error`, `saveSnapshot` throws instead of silently overwriting good values with a partial fetch. Walk extracted to `mergeSnapshotTree` and covered by tests.

25/25 tests pass. Ripple pass clean: no remaining `status?.stale ? 'text-gray-400'` grey-outs, no `?.global?.` reads bypassing `readGlobalField`, no direct `.frozen` access outside the helper.

PR #554, opened 2026-08-11
Add blog posts to sitemap, self-referencing canonicals, and pause Dev Rewards CTA

DavidMinarsch:
Four SEO/copy fixes from the site audit. No behavioural changes to any page beyond what is listed.

## 1. Blog posts in the sitemap

All 127 blog posts were missing from `sitemap.xml`. The sitemap is built by `next-sitemap` from the `pages/` directory, and posts are CMS-backed under `pages/blog/[id].tsx`, so it could not discover them. Combined with `/blog` rendering its list client-side, nothing gave a crawler a path to any post.

`additionalPaths` now pulls them from the CMS. Verified against production data: **127 posts, all unique, each with its publication date as `lastmod`**.

Two details worth noting for review:
- **Strapi caps page size at 100 regardless of `pagination[limit]`.** A single request asking for 1000 silently returns 100, so this pages through `meta.pagination.pageCount`.
- A CMS outage cannot fail the build — the fetch warns and returns `[]`, and the sitemap regenerates on the next deploy.

⚠️ **The same cap affects the site itself.** `getBlogs()` in `common-util/api/index.ts` requests `pagination[limit]: 1000` and therefore returns only the first 100 of 127 posts, so the blog index is silently missing 27. Not touched here — flagging for a separate fix.

## 2. Sitemap no longer advertises URLs that do not return 200

- `/academy` is redirected to `/404` in `next.config.js` but still had a page file, so it was listed. Excluded.
- Five slugs in `data/agents.json` redirect elsewhere (`prediction-agents`, `optimus`, `optimus-agent`, `modius-agent`, `mech`). Filtered out; agent paths go 17 → 12.
- `/restricted` excluded, and given a `noindex, nofollow` tag — a geo-block interstitial titled "Access Denied" should not be a search result.

`/deprecated-usecases` is deliberately left indexable.

## 3. Canonical URLs, and an `og:url` bug

`Meta.tsx` emitted **no canonical tag at all**, and hardcoded `og:url` and `twitter:url` to the site root on *every* page — so every share card pointed at the homepage rather than the page being shared.

Both now derive from `router.asPath`. `canonicalPath` can override where a page is a variant of another; `noindex` is opt-in. Uses `asPath` rather than `pathname` so blog posts get their own canonical instead of sharing one for `/blog/[id]`.

This also gives the topically-overlapping pages (`/agents/*` vs `/agent-economies/*`, and the three Mech pages) a self-referencing canonical each, so search engines stop having to guess. Their titles and descriptions are already distinct — no copy changes needed there.

## 4. Dev Rewards: paused, so no longer advertised as earnable

`/build` led with "Build on the Olas protocol and earn Dev Rewards" while a footnote read "*Dev Rewards program is temporarily not available*".

- Hero now leads with what is actually available.
- The card is retitled "Contribute code to the Olas Protocol", still explains what Dev Rewards is, and states plainly that the programme is paused — instead of an asterisk under a call to action.
- The "Learn more about Dev Rewards" trigger and its modal render are removed. `components/BuildPage/Content.tsx` is deliberately **kept in the repo**: the programme is paused, not cancelled, so restoring is re-adding one `showDevRewards` entry.

⚠️ One related line left alone: the Accelerator card still says "Have a chance to receive ongoing OLAS Dev Rewards for your registered agents." Whether that is still accurate depends on the Accelerator successor programme, which I do not have facts on.

### DIFF
--- a/components/Content/Articles.tsx
+++ b/components/Content/Articles.tsx
@@ -12,7 +12,36 @@ import Article from './Article';
 const API_URL = `${process.env.NEXT_PUBLIC_API_URL}/api`;
 const subURL = 'blog-posts';
 
-const fetcher = (url: string) => fetch(url).then((res) => res.json());
+/**
+ * Strapi caps a page at 100 entries whatever `pagination[limit]` asks for, so a
+ * single request for 1000 silently returns the first 100 and the rest of the
+ * blog simply never appears. Anything above the cap has to be paged through.
+ *
+ * Keep in sync with the equivalent paging in `next-sitemap.config.js`, which
+ * hits the same endpoint for the same reason.
+ */
+const STRAPI_MAX_PAGE_SIZE = 100;
+
+const fetcher = async ([url, maxPages]: [string, number]) => {
+  const first = await fetch(url).then((res) => res.json());
+
+  // `maxPages` keeps a small caller honest: the homepage asks for 3 posts, so
+  // it must not walk the whole archive just because more pages exist.
+  const pageCount = Math.min(first?.meta?.pagination?.pageCount ?? 1, maxPages);
+  if (pageCount <= 1) return first;
+
+  const rest = await Promise.all(
+    // Pages are 1-indexed and page 1 is already fetched.
+    Array.from({ length: pageCount - 1 }, (_, index) =>
+      fetch(`${url}&pagination[page]=${index + 2}`).then((res) => res.json())
+    )
+  );
+
+  return {
+    ...first,
+    data: [...(first?.data ?? []), ...rest.flatMap((page) => page?.data ?? [])],
+  };
+};
 
 const folders = [
   {
@@ -29,12 +58,16 @@ const Articles = ({ limit = 1000, showSeeAll = false, displayFolders, isMain }) 
   const params = {
     sort: ['datePublished:desc'],
     populate: '*',
-    'pagination[limit]': limit,
+    // `pageSize`, not `limit`: `limit` is silently clamped to the cap above and
+    // gives no `pageCount` to page through with.
+    'pagination[pageSize]': Math.min(limit, STRAPI_MAX_PAGE_SIZE),
   };
   const stringifyParams = qs.stringify(params);
+  const pageSize = Math.min(limit, STRAPI_MAX_PAGE_SIZE);
+  const maxPages = Math.ceil(limit / pageSize);
 
   const { data, isLoading } = useSWR(
-    `${API_URL}/${subURL}${params ? '?' : ''}${stringifyParams}`,
+    [`${API_URL}/${subURL}${params ? '?' : ''}${stringifyParams}`, maxPages],
     fetcher
   );
 
--- a/components/Meta.tsx
+++ b/components/Meta.tsx
@@ -1,6 +1,7 @@
 import { getLimitedText } from 'common-util/getLimitedText';
 import { getSiteUrl } from 'common-util/getSiteUrl';
 import Head from 'next/head';
+import { useRouter } from 'next/router';
 
 const TITLE_CHAR_MAX = 55;
 
@@ -18,6 +19,20 @@ type MetaProps = {
    * Use `''` for the home card (`/api/og`). Ignored when `siteImageUrl` is a non-empty URL.
    */
   ogPath?: string;
+  /**
+   * Overrides the canonical URL, which otherwise self-references the current
+   * path. Set this on a page that is a variant of another to point at the one
+   * that should rank.
+   */
+  canonicalPath?: string;
+  /** Keeps the page out of search indexes. For interstitials and error states. */
+  noindex?: boolean;
+};
+
+/** Strips the query string and trailing slash so canonicals stay stable. */
+const toCanonicalUrl = (siteUrl: string, path: string): string => {
+  const cleanPath = (path || '/').split('?')[0].split('#')[0].replace(/\/$/, '');
+  return cleanPath === '' ? siteUrl : `${siteUrl}${cleanPath}`;
 };
 
 const resolveShareImage = (
@@ -31,7 +46,19 @@ const resolveShareImage = (
   return SITE_DEFAULT_IMAGE_URL;
 };
 
-const Meta = ({ pageTitle, description, siteImageUrl, ogPath }: MetaProps) => {
+const Meta = ({
+  pageTitle,
+  description,
+  siteImageUrl,
+  ogPath,
+  canonicalPath,
+  noindex,
+}: MetaProps) => {
+  const router = useRouter();
+  // `asPath` is the real URL the visitor is on; `pathname` would keep the
+  // `[id]` placeholder and produce one shared canonical for every post.
+  const canonicalUrl = toCanonicalUrl(SITE_URL, canonicalPath ?? router?.asPath ?? '/');
+
   let title = pageTitle ? `${pageTitle} | ${SITE_TITLE}` : SITE_TITLE;
 
   if (title.length > TITLE_CHAR_MAX) {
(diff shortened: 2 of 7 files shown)


OjusWiZard (dismissed):
Approving — I checked the preview deployment's sitemap and it does what the description says: 176 entries, 127 blog posts, and `/academy`, `/restricted` and the five redirecting agent slugs are all gone. The `og:url` fix is a real win too; every page on prod currently points its share card at the homepage.

One question on section 2: was `/protocol` deliberately left in?

`next.config.js` 301s `/protocol` → `/stack`, but `pages/protocol.tsx` still exists, so filesystem discovery still picks it up — the preview sitemap has `https://olas.network/protocol` in it. That looks like the same case as `/academy` ("redirected to `/404` in `next.config.js` but still had a page file, so it was listed"), so I'd have expected it in `EXCLUDED_PATHS` alongside it.

Since you explicitly called out `/deprecated-usecases` as a deliberate exception, I didn't want to assume this one was an oversight. If it is, it's a one-line addition.

next-sitemap.config.js
```diff
@@ -1,16 +1,111 @@
...
+      // Fall back to the numeric id: `getBlog` resolves either, and a post with
+      // no slug is still better linked than not linked at all.
+      .map((post) => ({ slug: post?.slug || post?.id, date: post?.datePublished }))
+      .filter((post) => Boolean(post.slug))
+      .map((post) => ({
+        loc: `/blog/${post.slug}`,
+        changefreq: 'monthly',
+        // Posts are immutable once published, so the publication date is the
+        // honest lastmod. Omit it rather than send today's date.
+        ...(post.date ? { lastmod: new Date(post.date).toISOString() } : {}),
```
atepem, line 80: `new Date(post.date).toISOString()` throws a `RangeError` on an unparsable `datePublished` — and the throw lands in the outer `catch`, which returns `[]`. So a single malformed date on one post silently drops **every** blog URL from the sitemap, which is the exact outcome this function exists to prevent. The `console.warn` would say "could not fetch blog posts" even though the fetch succeeded.

Guard the parse per-post instead:

```js
const lastmod = post.date ? new Date(post.date) : null;
// ...
...(lastmod && !Number.isNaN(lastmod.getTime()) ? { lastmod: lastmod.toISOString() } : {}),
```

atepem (commented):
Review of the sitemap + canonical work. The copy/UI changes (`BuildPage/Hero.tsx`, `BuildPage/WaysToGrow.tsx`, `pages/restricted.tsx`) look good — the removed `showDevRewards`/`tip` fields have no remaining references, and leaving `Content.tsx` unimported is harmless.

Two things I'd want fixed before this merges, plus five smaller notes inline.

**1. `/blog` can now render empty where it previously rendered 100 posts.** See the inline note on `components/Content/Articles.tsx` — the `Promise.all` over later pages plus an unchecked `res.json()` means one failing request rejects the whole fetcher. That's a regression against `main`, which does a single request and always gets its 100.

**2. Blog posts still self-canonicalize at two URLs.** Not inline because the file isn't in the diff: `getBlog` in `common-util/api/index.ts` resolves *either* a numeric id or a slug, so every post is a 200 at both `/blog/123` and `/blog/my-slug`. With `<Meta>` now emitting a self-referencing canonical, each URL declares *itself* canonical and the duplication is entrenched rather than resolved. `pages/blog/[id].tsx` should pass an explicit ``canonicalPath={`/blog/${blog.slug}`}`` so both URLs point at the slug. This is the case the `canonicalPath` prop was added for, and right now it's the one place not using it.

For context on what I checked and found fine: next-sitemap 4.2.3 does dedupe `additionalPaths` against filesystem routes by relative `loc`, so agents that also have dedicated page files don't get double entries; and every dynamic page rendering `<Meta>` uses `getServerSideProps`, so `router.asPath` is correct in the server HTML (`/kits/[id]`, the one auto-static-optimized dynamic route, bails before rendering `<Meta>`).

next-sitemap.config.js
```diff
@@ -1,16 +1,111 @@
...
+    const posts = [];
+    let page = 1;
+    let pageCount = 1;
+
+    do {
+      const url = `${apiUrl}/api/blog-posts?sort[0]=datePublished:desc&pagination[page]=${page}&pagination[pageSize]=${PAGE_SIZE}`;
+      const response = await fetch(url);
+
+      if (!response.ok) {
+        throw new Error(`${response.status} ${response.statusText}`);
```
atepem, line 58: This `throw` escapes the `do/while` entirely, so a non-OK response on page 2 discards the 100 posts already collected from page 1 and the `catch` returns `[]`. The catch's comment promises graceful degradation, but the behaviour is all-or-nothing.

Since `posts` is already accumulated, `break` gives you the partial set the comment describes:

```js
if (!response.ok) {
  console.warn(`[sitemap] blog page ${page} failed (${response.status}) — using ${posts.length} posts fetched so far.`);
  break;
}
```

components/Content/Articles.tsx
```diff
@@ -12,7 +12,36 @@ import Article from './Article';
...
+
+const fetcher = async ([url, maxPages]: [string, number]) => {
+  const first = await fetch(url).then((res) => res.json());
+
+  // `maxPages` keeps a small caller honest: the homepage asks for 3 posts, so
+  // it must not walk the whole archive just because more pages exist.
+  const pageCount = Math.min(first?.meta?.pagination?.pageCount ?? 1, maxPages);
+  if (pageCount <= 1) return first;
+
+  const rest = await Promise.all(
```
atepem, line 33: Two failure modes here, and together they make `/blog` worse than it is on `main`:

1. `Promise.all` rejects on the first failing page, so one flaky request for page 2 throws away page 1's 100 posts and SWR renders the error state — an empty index. Today's single-request code would still have shown 100.
2. `res.json()` is unchecked. A Strapi 500 with an HTML error body throws inside the `.then`; a 500 with a JSON error body resolves to an object with no `data`, which `?? []` quietly swallows.

`Promise.allSettled` plus an `ok` check keeps whatever pages did come back:

```ts
const rest = await Promise.allSettled(
  Array.from({ length: pageCount - 1 }, (_, index) =>
    fetch(`${url}&pagination[page]=${index + 2}`).then((res) => {
      if (!res.ok) throw new Error(`${res.status} ${res.statusText}`);
      return res.json();
    })
  )
);

return {
  ...first,
  data: [
    ...(first?.data ?? []),
    ...rest.flatMap((r) => (r.status === 'fulfilled' ? r.value?.data ?? [] : [])),
  ],
};
```

Worth applying the same `ok` check to the `first` fetch too.

components/Content/Articles.tsx
```diff
@@ -29,12 +58,16 @@ const Articles = ({ limit = 1000, showSeeAll = false, displayFolders, isMain })
   const params = {
     sort: ['datePublished:desc'],
```
atepem, line 59: Paging is only safe if the sort key is a total order, and `datePublished` is day-precision — several posts published the same day tie. Strapi/Postgres gives no stable order among ties across separate queries, so a tie straddling the page-100 boundary can return one post on both pages and drop another entirely. Silent, and it only shows up once the archive crosses 100.

Add a unique tiebreaker:

```ts
sort: ['datePublished:desc', 'id:desc'],
```

Same fix needed in the two other new paging sites — `getBlogs` in `common-util/api/index.ts` and the `sort[0]=datePublished:desc` URL in `next-sitemap.config.js`.

next-sitemap.config.js
```diff
@@ -1,16 +1,111 @@
 const agentsData = require('./data/agents.json');
 
+/**
+ * Agent slugs in `data/agents.json` that `next.config.js` redirects elsewhere.
+ * They must not be listed in the sitemap: a sitemap should only contain URLs
+ * that return 200, and a redirecting entry wastes crawl budget and dilutes the
+ * signal for the page it redirects to.
+ */
+const REDIRECTED_AGENT_SLUGS = new Set([
```
atepem, line 9: This is a third place encoding "which agent slugs are dead", after `removePage` in `data/agents.json` and the redirect table in `next.config.js`. The next agent retired via `removePage: true` alone will be advertised in the sitemap as a URL that 404s — the precise failure the comment above says this list exists to prevent.

Since `agents.json` already carries the flag, deriving it removes the drift:

```js
const agentsPaths = agentsData
  .filter((agent) => !agent.removePage && !REDIRECTED_AGENT_SLUGS.has(agent.slug))
  .map(/* ... */);
```

Keeping the explicit set as a belt-and-braces for slugs that redirect *without* `removePage` is fine — the point is that `removePage` alone should be sufficient.

components/Meta.tsx
```diff
@@ -18,6 +19,20 @@ type MetaProps = {
    * Use `''` for the home card (`/api/og`). Ignored when `siteImageUrl` is a non-empty URL.
    */
   ogPath?: string;
+  /**
+   * Overrides the canonical URL, which otherwise self-references the current
+   * path. Set this on a page that is a variant of another to point at the one
+   * that should rank.
+   */
+  canonicalPath?: string;
```
atepem, line 27: `canonicalPath` and `noindex` are both added but no caller passes either, so as merged this is dead API. Specifically:

- `noindex`: `/restricted` hand-rolls its own `<meta name="robots">` in this same PR rather than rendering `<Meta noindex />`, so the two mechanisms already disagree about where that decision lives.
- `canonicalPath`: the variant-page consolidation the PR description cites never happens — see the blog id-vs-slug case in my summary comment, which is the clearest caller for it.

Not arguing to drop the props; wiring up the blog canonical would give `canonicalPath` its first real use and make the prop's purpose self-evident.

PR #2, opened 2023-06-27
chore: minor fixes

DavidMinarsch:

PR #555, opened 2026-08-11
fix(verify): parity script — cursor pagination + displayable-rows redefinition + analytics dedup

DIvyaNautiyal07:
## Summary

Four fixes to `scripts/verify-mech-analytics-parity.mjs` so the parity artifact produces a clean PASS on all six checks (2 platforms × 3 checks) against production. Local run on 2026-08-07 → 2026-08-11 shows all PASS.

## Fixes

**1. Cursor-based subgraph pagination** (commit `41c66c7f`, refined in `f2f7143e`).

The previous `first: 1000, skip: X` with `SUBGRAPH_MAX_SKIP = 5000` capped every fetch at 6,000 rows. At ~21K events/day on Gnosis, even a 1-day window overflowed the cap and reported truncation as a coverage gap. Switched to cursor pagination on `blockTimestamp` (with a `_gte` + id-based dedup for tie-boundary correctness — see fix #4 below).

**2. Analytics-side hex/decimal dedup** (commit `c6141d4d`).

Mech-analytics's `per_request_scores` currently stores some request_ids in hex form (`0x…`) and some in decimal form (uint256 stringified). Same on-chain request appears as two rows under different keys. Verified: for one requester on omenstrat in a 4-day window, 359 raw rows collapsed to 241 unique after normalising `(requester, request_id)`. This dedup is a safety net until the predict-api canonicalisation PR (valory-xyz/predict-api#178) merges and its data migration runs — after that this dedup becomes a no-op but stays as a guard against regressions.

**3. `row_parity` redefinition: compare "displayable rows" instead of "all rows"** (commit `f2f7143e`).

The earlier definition (all subgraph rows vs all analytics rows) was comparing two different populations by accident, because pending-delivery and shell rows are represented differently on each side. Redefined to compare the population that actually renders on the predict page as a completed prediction:

- Subgraph filter: `isDelivered: true AND parsedRequest.questionTitle IS NOT NULL` (empty-string title also excluded; there are prod rows with `questionTitle = ""`).
- Analytics side: `/v1/data/scored-rows` only (drops the union with `/v1/data/unscored-rows` we added earlier — that was solving the wrong problem).

Everything else drops symmetrically on both sides:
- Undelivered rows: `isDelivered: false` on subgraph → excluded. Not in scored-rows on analytics → excluded.
- Shell rows: no `questionTitle` on subgraph → excluded. Not in scored-rows on analytics → excluded.
- Off-chain: only surfaced on analytics via unscored; excluded from row_parity by construction now.

The off-chain (`fetchMarketplaceOffChainRequestIds`) and unscored (`fetchUnscoredRows`) fetches are kept in the codebase for observability — their counts appear in the artifact as `subgraph_off_chain_count` and `analytics_unscored_count` so an operator can see the composition, but they do NOT feed the diff.

**4. Tie-boundary pagination bug fix** (commit `f2f7143e`).

`_gt: lastTs` per page drops rows when more than one row shares the last row's `blockTimestamp` and not all of them fit on the current page. Switched to `_gte: lastTs` + id-based dedup across page boundaries. Verified on omenstrat: `_gt` returned 6,392 rows; `_gte + dedup` returns 6,393 (one tie-boundary row was previously showing as a false "extra" on the analytics side).

## Verdict (2026-08-07 → 2026-08-11)

| Platform | Check | Result |
|---|---|---|
| omenstrat | open_market_count | PASS (6122 = 6122) |
| omenstrat | row_parity | **PASS** (6393 = 6393, 0 missing / 0 extra) |
| omenstrat | sender_total | PASS (302,465 = 302,465) |
| polystrat | open_market_count | PASS (383 = 383) |
| polystrat | row_parity | **PASS** (383 = 383, 0 missing / 0 extra) |
| polystrat | sender_total | PASS (64,983 = 64,983) |

Local test suite: all 83 lib unit tests pass. ESLint clean.

## Related

- Depends on nothing to merge (works today) but pairs cleanly with:
  - valory-xyz/predict-api#178 — canonicalises `request_id` encoding at both writers + data migration. Once that ships, the dedup in fix #2 becomes structurally redundant (kept as regression guard).
  - valory-xyz/mech-analytics#28 — adds `source` column to `per_request_scores` + bundled canonicaliser script for existing rows.

## Test plan

- [x] Local run against production DBs shows all 6 checks PASS
- [x] `node --test scripts/verify-mech-analytics-parity.test.mjs` — all 83 tests pass
- [x] ESLint clean

### DIFF
--- a/scripts/verify-mech-analytics-parity.mjs
+++ b/scripts/verify-mech-analytics-parity.mjs
@@ -727,6 +727,125 @@ const fetchDailyStats = async (url, dateGte, dateLte, participantTitleField, lab
   return { stats, truncated };
 };
 
+// Off-chain request IDs live only on `MarketplaceDeliveryWithSignatures`
+// entities — the off-chain handler
+// (autonolas-subgraph/subgraphs/marketplace/src/marketplace/mech-marketplace.ts
+// handleMarketplaceDeliveryWithSignatures) does NOT create a `Request`
+// entity per off-chain request. Instead it emits one delivery entity
+// per settled batch, carrying a `requestIds` array of the off-chain IDs
+// that batch settled. Without this the row-parity check 2 double-counts
+// those rows as "extra in analytics" (they land in
+// /v1/data/scored-rows|unscored-rows via the predict-api lake, but the
+// subgraph's `requests(...)` query never sees them).
+//
+// Cursor-paginated on `blockTimestamp` — same shape and truncation /
+// same-timestamp handling as `fetchMarketplaceRequests` above. The
+// entity's own `id` is the batch id (uses transactionHash + log index
+// under the hood, `130 chars` shape) and is NOT the per-request id we
+// want; the per-request ids are the entries inside the `requestIds`
+// array, one entity yielding N entries where N = numDeliveries. Each
+// entry is normalised via `normaliseRequestId` (subgraph returns lower-
+// case hex Bytes; analytics stores the same uint256 as a decimal string
+// for marketplace-era rows; normaliser collapses both to decimal so
+// the id-based multiset diff is format-agnostic).
+const fetchMarketplaceOffChainRequestIds = async (url, windowStartSec, windowEndSec, label) => {
+  const entities = [];
+  let cursor = windowStartSec;
+  let truncated = false;
+  let pages = 0;
+  for (;;) {
+    const data = await gqlRequest(
+      url,
+      `query MarketplaceOffChainDeliveries {
+        marketplaceDeliveryWithSignatures_collection(
+          first: ${SUBGRAPH_PAGE}
+          where: { blockTimestamp_gt: "${cursor}", blockTimestamp_lte: "${windowEndSec}" }
+          orderBy: blockTimestamp
+          orderDirection: asc
+        ) {
+          id
+          requester
+          numDeliveries
+          requestIds
+          blockTimestamp
+        }
+      }`,
+      label
+    );
+    const page = data?.marketplaceDeliveryWithSignatures_collection ?? [];
+    if (verbose)
+      emit(`  [${label}] deliveries page ${pages + 1} cursor=${cursor} rows=${page.length}`);
+    entities.push(...page);
+    pages += 1;
+    if (page.length < SUBGRAPH_PAGE) break;
+    if (pages >= SUBGRAPH_MAX_PAGES) {
+      truncated = true;
+      break;
+    }
+    const lastTs = Number(page[page.length - 1].blockTimestamp);
+    const firstTs = Number(page[0].blockTimestamp);
+    if (lastTs === firstTs) {
+      emit(
+        `  [${label}] WARN: same-timestamp saturation at ts=${lastTs} (full page shares a single blockTimestamp) — advancing cursor by 1s`
+      );
+      cursor = lastTs + 1;
+    } else {
+      cursor = lastTs;
+    }
+  }
+
+  const withRequestId = [];
+  let skippedMissingSenderOrTs = 0;
+  let skippedEmptyRequestIds = 0;
+  let skippedAfterWindowEnd = 0;
+  let skippedUnparseableId = 0;
+  for (const entity of entities) {
+    // `requester` on the delivery entity is a Bytes column (not a
+    // `Sender` reference), so no `.id` accessor is needed. Lower-case
+    // to match the subgraph's Request-side normalisation (`sender.id`
+    // is lower-cased by convention).
+    const requester = entity.requester?.toLowerCase();
+    const ts = Number(entity.blockTimestamp ?? 0);
+    const ids = Array.isArray(entity.requestIds) ? entity.requestIds : [];
+    if (!requester || ts <= 0) {
+      skippedMissingSenderOrTs += 1;
+      continue;
+    }
+    if (ts > windowEndSec) {
+      skippedAfterWindowEnd += 1;
+      continue;
+    }
+    if (ids.length === 0) {
+      skippedEmptyRequestIds += 1;
+      continue;
+    }
+    for (const raw of ids) {
+      const requestId = normaliseRequestId(raw);
+      if (!requestId) {
+        skippedUnparseableId += 1;
+        continue;
+      }
+      // No title on off-chain rows — the delivery entity doesn't carry
+      // a parsed title (title lives in the private IPFS payload, which
+      // the subgraph never sees). Analytics-side rows for these same
+      // requests also have `question_title` NULL for the same reason.
+      // Set title=null; the id-based match in diffRowSets is what
+      // pairs them.
+      withRequestId.push({ requester, ts, title: null, requestId });
+    }
+  }
+
+  return {
+    withRequestId,
+    truncated,
+    entitiesFetched: entities.length,
+    skippedMissingSenderOrTs,
+    skippedEmptyRequestIds,
+    skippedAfterWindowEnd,
+    skippedUnparseableId,
+  };
+};
+
 // getMarketplaceSendersQuery (queries.ts) narrowed to the Safes under
 // verification via id_in — the site pages every sender, but check 3
 // only needs the registered Safes' counters. See pickSubgraphSenderTotal
@@ -871,37 +990,55 @@ try {
     const dailyStatsUrl = process.env[platform.dailyStatsUrlEnv];
 
     section(`${key} (chain ${chainId}, agent ${agentId}) — data pulls`);
-    const [marketplace, dailyStats, scoredRows, unscoredRows, aggregate] = await Promise.all([
-      fetchMarketplaceRequests(marketplaceUrl, windowStartSec, windowEndSec, `${key}:marketplace`),
-      // date_lte is capped at (now - finality lag) so the subgraph
-      // side doesn't consume settlements still inside the analytics
-      // 24h finality gate — otherwise recent settlements would show as
-      // consumed on the subgraph but pending in analytics and produce
-      // a chronic delta on honest runs. `nowSec` for the lag is
-      // computed against the actual clock, not the window end, so an
-      // explicit --since/--until run still trims the DailyStats side
-      // by the same gate.
-      fetchDailyStats(
-        dailyStatsUrl,
-        Math.floor(windowStartSec / DAY_SECONDS) * DAY_SECONDS,
-        // For a trailing window the cap is (now - finality lag) — the
-        // usual guard. For an explicit --since/--until historical window
-        // the cap becomes (windowEnd + finality lag), so we don't
-        // uselessly sweep months of daily-stats to consume settlements
-        // that already happened well after the window closed. Both
-        // stay in effect (`min`), so operators can't accidentally
-        // widen the consumption horizon by picking a huge window.
-        Math.min(
-          Math.floor(Date.now() / 1000) - finalityLagHours * 60 * 60,
-          windowEndSec + finalityLagHours * 60 * 60
+    const [marketplace, marketplaceOffChain, dailyStats, scoredRows, unscoredRows, aggregate] =
+      await Promise.all([
+        fetchMarketplaceRequests(
+          marketplaceUrl,
+          windowStartSec,
+          windowEndSec,
+          `${key}:marketplace`
         ),
-        platform.participantTitleField,
-        `${key}:daily-stats`
-      ),
-      fetchScoredRows(chainId, windowStartSec, windowEndSec, `${key}:scored-rows`),
-      fetchUnscoredRows(chainId, windowStartSec, windowEndSec, `${key}:unscored-rows`),
-      fetchAgentAggregate(chainId, agentId),
-    ]);
+        // Off-chain request IDs from `MarketplaceDeliveryWithSignatures`
+        // entities — the subgraph's `requests(...)` query only returns
+        // on-chain rows (the off-chain handler doesn't create `Request`
+        // entities). Unioned into the subgraph side of check 2 below so
+        // the two sides model the same superset. See
+        // `fetchMarketplaceOffChainRequestIds` for the rationale.
+        fetchMarketplaceOffChainRequestIds(
+          marketplaceUrl,
+          windowStartSec,
+          windowEndSec,
+          `${key}:marketplace-offchain`
+        ),
+        // date_lte is capped at (now - finality lag) so the subgraph
+        // side doesn't consume settlements still inside the analytics
+        // 24h finality gate — otherwise recent settlements would show as
+        // consumed on the subgraph but pending in analytics and produce
+        // a chronic delta on honest runs. `nowSec` for the lag is
+        // computed against the actual clock, not the window end, so an
+        // explicit --since/--until run still trims the DailyStats side
+        // by the same gate.
+        fetchDailyStats(
+          dailyStatsUrl,
+          Math.floor(windowStartSec / DAY_SECONDS) * DAY_SECONDS,
+          // For a trailing window the cap is (now - finality lag) — the
+          // usual guard. For an explicit --since/--until historical window
+          // the cap becomes (windowEnd + finality lag), so we don't
+          // uselessly sweep months of daily-stats to consume settlements
+          // that already happened well after the window closed. Both
+          // stay in effect (`min`), so operators can't accidentally
+          // widen the consumption horizon by picking a huge window.
+          Math.min(
+            Math.floor(Date.now() / 1000) - finalityLagHours * 60 * 60,
+            windowEndSec + finalityLagHours * 60 * 60
+          ),
+          platform.participantTitleField,
+          `${key}:daily-stats`
+        ),
+        fetchScoredRows(chainId, windowStartSec, windowEndSec, `${key}:scored-rows`),
+        fetchUnscoredRows(chainId, windowStartSec, windowEndSec, `${key}:unscored-rows`),
+        fetchAgentAggregate(chainId, agentId),
+      ]);
     emit(`  marketplace requests in window (QMR-usable): ${marketplace.rows.length}`);
     emit(
       `    (skipped: missing sender/ts=${marketplace.skippedMissingSenderOrTs}, ` +
@@ -911,6 +1048,17 @@ try {
     );
     emit(
       `  marketplace rows with request_id (row-parity feed): ${marketplace.withRequestId.length}`
+    );
+    emit(
+      `  marketplace off-chain request_ids (from ${marketplaceOffChain.entitiesFetched} ` +
+        `MarketplaceDeliveryWithSignatures entities): ${marketplaceOffChain.withRequestId.length}` +
+        `${marketplaceOffChain.truncated ? ' (TRUNCATED at subgraph page cap)' : ''}`
+    );
+    emit(
+      `    (delivery-entity skips: missing sender/ts=${marketplaceOffChain.skippedMissingSenderOrTs}, ` +
+        `empty requestIds=${marketplaceOffChain.skippedEmptyRequestIds}, ` +
+        `after window end=${marketplaceOffChain.skippedAfterWindowEnd}, ` +
+        `unparseable id=${marketplaceOffChain.skippedUnparseableId})`
     );
     emit(
       `  daily-stat rows: ${dailyStats.stats.length}` +
@@ -1095,9 +1243,23 @@ try {
 
     // ---- Check 2: scored-rows row parity ----------------------------------
     section(`${key} — check 2: scored-rows row parity`);
-    const rowDiff = diffRowSets(marketplace.withRequestId, analyticsRowsForCheck2);
+    // Subgraph side is the union of on-chain requests and off-chain
+    // request IDs. On-chain rows come from the `requests(...)` query
+    // (handleMarketplaceRequest creates one `Request` entity per event);
+    // off-chain rows only surface as entries inside
+    // `MarketplaceDeliveryWithSignatures.requestIds` because the off-
+    // chain handler doesn't create per-request entities. Analytics
+    // side is scored ∪ unscored, since predict-api writes both on-chain
+    // and off-chain rows to the same lake. Without the off-chain leg
+    // every off-chain row shows up as "extra in analytics" (~4,600 on
+    // omenstrat, ~360 on polystrat in a 4-day window, verified 2026-08).
+    const subgraphOnChainRows = marketplace.withRequestId;
+    const subgraphOffChainRows = marketplaceOffChain.withRequestId;
+    const subgraphAllRows = [...subgraphOnChainRows, ...subgraphOffChainRows];
+    const rowDiff = diffRowSets(subgraphAllRows, analyticsRowsForCheck2);
     emit(
-      `  subgraph rows: ${marketplace.withRequestId.length}, analytics rows: ${analyticsRowsForCheck2.length} ` +
+      `  subgraph rows: ${subgraphAllRows.length} (on-chain=${subgraphOnChainRows.length} + ` +
+        `off-chain=${subgraphOffChainRows.length}), analytics rows: ${analyticsRowsForCheck2.length} ` +
         `(scored=${scoredRowsInParity} + unscored=${unscoredRowsInParity})`
     );
     emit(`  missing in analytics: ${rowDiff.missingInAnalytics}`);
@@ -1113,9 +1275,13 @@ try {
     for (const sample of rowDiff.missingSamples) emit(`    missing: ${sample}`);
     for (const sample of rowDiff.extraSamples) emit(`    extra:   ${sample}`);
 
-    const check2Truncated = marketplace.truncated || scoredRows.truncated || unscoredRows.truncated;
+    const check2Truncated =
+      marketplace.truncated ||
+      marketplaceOffChain.truncated ||
+      scoredRows.truncated ||
+      unscoredRows.truncated;
     const { status: check2Status, reason: check2Reason } = resolveCheck2Status({
-      subgraphRowCount: marketplace.withRequestId.length,
+      subgraphRowCount: subgraphAllRows.length,
       analyticsRowCount: analyticsRowsForCheck2.length,
       missingInAnalytics: rowDiff.missingInAnalytics,
       extraInAnalytics: rowDiff.extraInAnalytics,
@@ -1137,7 +1303,12 @@ try {
       platform: key,
       status: check2Status,
       status_reason: check2Reason,
-      subgraph_rows: marketplace.withRequestId.length,
+      // Composition: `subgraph_rows` is the union used to compare;
+      // the on-chain / off-chain split is preserved so an operator can
+      // see which leg drives a residual delta.
+      subgraph_rows: subgraphAllRows.length,
+      subgraph_on_chain_rows: subgraphOnChainRows.length,
+      subgraph_off_chain_rows: subgraphOffChainRows.length,
       subgraph_qmr_rows: marketplace.rows.length,
       analytics_rows: analyticsRowsForCheck2.length,
       analytics_scored_rows: scoredRowsInParity,
--- a/scripts/verify-mech-analytics-parity.mjs
+++ b/scripts/verify-mech-analytics-parity.mjs
@@ -39,12 +39,31 @@
 //   divergence in prod. The (a)/(b)/(c) row-bucket breakdown is still
 //   reported for context but is not the gated number.
 //
-// Check 2 — scored-rows row parity.
-//   For the same window, the row set from
-//   getMechRequestsIncrementalQuery (requester, blockTimestamp,
-//   questionTitle) must match the rows from /v1/data/scored-rows
-//   (requester, requested_at, question_title). Missing / extra rows
-//   are reported per side.
+// Check 2 — row parity ("displayable rows" definition).
+//   For the same window, the set of rows a user would see rendered
+//   on the predict page as a completed prediction must match on both
+//   sides. Symmetric filter:
+//     * subgraph side: `isDelivered: true AND
+//       parsedRequest.questionTitle IS NOT NULL` in the WHERE clause
+//       of the marketplace-subgraph `requests` query.
+//     * analytics side: `/v1/data/scored-rows` ONLY (scored rows have
+//       `tool NOT NULL AND delivered_at NOT NULL` by construction, so
+//       they are exactly the delivered + parseable set).
+//   Off-chain requests (subgraph `MarketplaceDeliveryWithSignatures`)
+//   and unscored/shell rows (analytics `/v1/data/unscored-rows`)
+//   drop symmetrically on both sides — they are not in the
+//   displayable-rows population. Their counts are surfaced in the
+//   artifact as `subgraph_off_chain_count` and
+//   `analytics_unscored_count` for observability only.
+//   The prior definition compared "all subgraph requests" against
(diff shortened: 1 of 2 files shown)


scripts/verify-mech-analytics-parity.mjs
```diff
@@ -553,18 +784,158 @@ const fetchDailyStats = async (url, dateGte, dateLte, participantTitleField, lab
...
     }
+    const lastDate = Number(page[page.length - 1].date);
+    const firstDate = Number(page[0].date);
+    if (lastDate === firstDate) {
+      emit(
+        `  [${label}] WARN: same-date saturation at date=${lastDate} (full page shares a single date) — advancing cursor by 1 day`
+      );
+      cursor = lastDate + DAY_SECONDS;
+    } else {
+      cursor = lastDate;
```
Tanya-atatakai, line 791: please let's apply the `_gte` + dedup fix here too — this is the same tie-boundary bug as fix #4, unfixed. the normal advance is `cursor = lastDate` with `date_gt: ${cursor}` (line 775), so when a page boundary splits inside one date (the common case: hundreds of traders share every `date`, so any multi-page pull splits mid-date), the remaining rows at `lastDate` are silently dropped. that corrupts check 1's consumption math while looking green, and the header comment explicitly supports months-long `--since/--until` runs where multi-page is guaranteed. mirror `fetchMarketplaceRequests`: add `id` to the selection, switch to `date_gte` after the first page, dedup by entity id.

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -553,18 +784,158 @@ const fetchDailyStats = async (url, dateGte, dateLte, participantTitleField, lab
...
       truncated = true;
       break;
     }
+    const lastDate = Number(page[page.length - 1].date);
+    const firstDate = Number(page[0].date);
+    if (lastDate === firstDate) {
+      emit(
+        `  [${label}] WARN: same-date saturation at date=${lastDate} (full page shares a single date) — advancing cursor by 1 day`
+      );
+      cursor = lastDate + DAY_SECONDS;
```
Tanya-atatakai: the saturation advance skips a full day: `date_gt: lastDate + DAY_SECONDS` returns `date >= lastDate + 2 days`, so the entire day after the saturated one silently vanishes. the `_gt` itself is the advance:

```suggestion
      cursor = lastDate; // `date_gt` below already skips the saturated date
```

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -429,119 +461,318 @@ for (const name of REQUIRED_ENV.slice(1)) {
...
       break;
     }
+    const lastTs = Number(page[page.length - 1].blockTimestamp);
+    const firstTs = Number(page[0].blockTimestamp);
+    if (lastTs === firstTs) {
+      emit(
+        `  [${label}] WARN: same-timestamp saturation at ts=${lastTs} (full page shares a single blockTimestamp) — advancing cursor by 1s`
+      );
+      cursor = lastTs + 1;
+      firstPage = true; // reuse `_gt` on the leap; no boundary to overlap
```
Tanya-atatakai: same off-by-one here: `cursor = lastTs + 1` + `_gt` means the next page starts at `ts >= lastTs + 2`, so rows at exactly `lastTs + 1` are lost — one second more than the acknowledged loss the WARN describes.

```suggestion
      cursor = lastTs; // the `_gt` leap below IS the +1s advance
      firstPage = true;
```

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -553,18 +784,158 @@ const fetchDailyStats = async (url, dateGte, dateLte, participantTitleField, lab
...
+      truncated = true;
+      break;
+    }
+    const lastTs = Number(page[page.length - 1].blockTimestamp);
+    const firstTs = Number(page[0].blockTimestamp);
+    if (lastTs === firstTs) {
+      emit(
+        `  [${label}] WARN: same-timestamp saturation at ts=${lastTs} (full page shares a single blockTimestamp) — advancing cursor by 1s`
+      );
+      cursor = lastTs + 1;
```
Tanya-atatakai: and same here in the off-chain fetcher — plus the header comment at line 831 says "same shape and truncation / same-timestamp handling as `fetchMarketplaceRequests`" but this loop is `_gt`-only with no id dedup, so the tie-boundary fix was not actually carried over. observability-only, so lower stakes, but can we either apply the same `_gte` + dedup or fix the comment so it stops promising it? as written the next reader will trust it.

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -429,119 +461,318 @@ for (const name of REQUIRED_ENV.slice(1)) {
...
       url,
       `query MechRequestsIncremental {
         requests(
           first: ${SUBGRAPH_PAGE}
-          skip: ${skip}
-          where: { blockTimestamp_gt: "${windowStartSec}" }
+          where: {
+            ${tsPredicate},
+            blockTimestamp_lte: "${windowEndSec}",
+            isDelivered: true,
```
Tanya-atatakai: a bit confused about check 1's fidelity after this change: its charter says "same QMR + dailyStats consumption pipeline the consumer runs", but the site's `fetchIncrementalMechRequests` doesn't filter `isDelivered`, and the site's `iterateScoredRows` doesn't pass `source` — both filters are new here and shared with check 2 b/c one query/fetch feeds both checks. today it's a no-op (off-chain = 0, in-flight ≈ 0 under the 24h lag), but post-switch check 1 stops modelling what the site actually computes. intentional? if yes, maybe document the divergence at the check-1 section; if no, the check-1 feed needs its own unfiltered query. wdyt

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -682,53 +1117,197 @@ try {
...
+        // fires on on-chain Request events), so filtering here makes the
+        // subgraph-vs-analytics comparison symmetric at the API
+        // boundary rather than by incidental drop-out downstream. Off-
+        // chain rows (`mech_offchain`) don't appear in the subgraph's
+        // `Request` collection at all — they're separately fetched via
+        // `fetchMarketplaceOffChainRequestIds` for observability only.
+        // The API is defined by `SourceLabel` in mech-analytics
+        // `etl/api/routes/data.py` (PR #28); typos return 422.
+        fetchScoredRows(chainId, windowStartSec, windowEndSec, `${key}:scored-rows`, {
+          source: ['mech_onchain', 'ipfs_historical'],
```
Tanya-atatakai, line 1185: jsyk — the `?source=` filter and `/v1/data/unscored-rows` only exist once mech-analytics #27/#28 are deployed. against an older deployment FastAPI silently ignores unknown query params, so the scored pull would quietly include off-chain rows and parity would compare the wrong population while looking healthy (the "typos return 422" comment only holds post-#28). worth a cheap startup probe (request a bogus `source` and require the 422) so the script hard-fails on an old deployment instead?

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -429,119 +461,318 @@ for (const name of REQUIRED_ENV.slice(1)) {
...
+// string. Predict-api stores mech_requests.request_id as the same
+// 256-bit value in decimal (from the on-chain event's ``requestId``
+// uint256) for marketplace-era rows, and mech-analytics carries that
+// through unchanged — but pre-marketplace rows in per_request_scores
+// (2023-era backfill on chain 100 and some rollovers on chain 137)
+// were stored as 0x-hex too. Result: the analytics API mixes both
+// formats depending on when the row was written. Normalising both
+// sides to the same base10 representation makes id-based parity
+// order-invariant and format-agnostic.
+const normaliseRequestId = (raw) => {
```
Tanya-atatakai, line 468: can we move `normaliseRequestId` (and ideally the cursor-advance decision — "given a page, what's the next predicate + cursor") into the lib and cover them with tests? all the new logic in this PR lives in the untested main script, the 83 lib tests didn't grow, and the two saturation off-by-ones flagged above are exactly the kind of thing a 10-line unit test catches.

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -429,119 +461,318 @@ for (const name of REQUIRED_ENV.slice(1)) {
...
+// ``blockTimestamp_lte: windowEndSec`` puts the upper bound in the query
+// so late rows don't need a post-fetch filter. Same query shape as
 // getMechRequestsIncrementalQuery (common-util/graphql/queries.ts:816)
-// + fetchIncrementalMechRequests (roi-distribution.ts:229). A full page
-// at the graph-node skip ceiling marks the result truncated instead of
-// silently under-counting.
+// + fetchIncrementalMechRequests (roi-distribution.ts:229) otherwise.
+//
+// Row-parity population — "displayable rows": the WHERE clause filters
+// to `isDelivered: true AND parsedRequest_: {questionTitle_not: ""}`.
```
Tanya-atatakai: tiny, the "displayable rows" rationale essay appears near-identically in four places (this header, `fetchUnscoredRows`, the check-2 section, `buildAnalyticsRowParityEntry`). up to you, but I'd keep one canonical block and point the others at it — four copies will drift.

Tanya-atatakai (changes requested):
The core `_gte` + id-dedup pagination fix is correct and the 6392→6393 verification is exactly the right evidence; the `questionTitle_not: ""` claim checks out (SQL `NULL != ''` is NULL → null titles excluded, plus the nested filter requires `parsedRequest` non-null); the analytics-side `(requester, request_id)` dedup Map also absorbs scored-rows' mid-pagination re-serves, which would otherwise count as extras; and keeping GraphQL-level errors fatal while retrying only 5xx/network is the right split.

But tracing the cursor arithmetic in all three fetchers found the same bug class this PR set out to eliminate, still live in `fetchDailyStats` (inline comments, most-severe first).

**Verdict: needs changes** — the `fetchDailyStats` tie-boundary loss silently corrupts check 1's consumption math, and it feeds a gated check.

### What I verified

- read the lib's `diffRowSets` at the PR head — the new `{requester, row}` entry shape matches its existing contract, and with both sides now id-gated the ts-fallback path goes dormant (fine)
- recomputed the cursor arithmetic for all three fetchers by hand — the three off-by-one/tie findings are from that trace, not speculation
- cross-checked the check-1 charter against the site's `fetchIncrementalMechRequests` and `iterateScoredRows` (neither filters `isDelivered`/`source`)
- confirmed `dailyProfitStatistics` fetches no `id`, so no dedup is even possible as written
- did not run the script against production (needs prod env keys + hits live APIs); the PASS table is the author's run

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -429,119 +461,318 @@ for (const name of REQUIRED_ENV.slice(1)) {
...
       `query MechRequestsIncremental {
         requests(
           first: ${SUBGRAPH_PAGE}
-          skip: ${skip}
-          where: { blockTimestamp_gt: "${windowStartSec}" }
+          where: {
+            ${tsPredicate},
+            blockTimestamp_lte: "${windowEndSec}",
+            isDelivered: true,
+            parsedRequest_: { questionTitle_not: "" }
```
OjusWiZard: **Important — this also silently redefines check 1, not just check 2.**

`fetchMarketplaceRequests` feeds *both* checks, so adding `isDelivered: true` here changed check 1's subgraph population too — but `f2f7143e`'s stated scope is the row_parity redefinition.

The live consumer applies no such filter:
- `getMechRequestsIncrementalQuery` (`common-util/graphql/queries.ts:824-842`) filters on `blockTimestamp_gt` only.
- `fetchIncrementalMechRequests` (`roi-distribution.ts:246-253`) admits any row with `agentId && questionTitle && ts > 0`.

So **undelivered-but-titled requests do enter the live QMR**, and the pre-PR script matched that exactly (`where: { blockTimestamp_gt: ... }`). They now drop from the subgraph side, while the analytics side (scored => delivered by construction) never had them. A genuine post-swap divergence in that population is no longer detectable by any check — and the header at L23-33 still claims check 1 runs "the exact algorithm roi-distribution.ts runs today".

Suggest either fetching check 1's feed unfiltered (restoring consumer fidelity) or updating the header to state check 1 now gates the delivered-only subset.

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -429,119 +461,318 @@ for (const name of REQUIRED_ENV.slice(1)) {
...
       truncated = true;
       break;
     }
+    const lastTs = Number(page[page.length - 1].blockTimestamp);
+    const firstTs = Number(page[0].blockTimestamp);
+    if (lastTs === firstTs) {
+      emit(
+        `  [${label}] WARN: same-timestamp saturation at ts=${lastTs} (full page shares a single blockTimestamp) — advancing cursor by 1s`
+      );
+      cursor = lastTs + 1;
```
OjusWiZard: **Important — off-by-one: this skips an entire extra second.**

`cursor = lastTs + 1` combined with `firstPage = true` makes the next predicate `blockTimestamp_gt: "${lastTs + 1}"`, which excludes every row at `ts == lastTs + 1` on top of the intended `ts == lastTs` sacrifice. Demonstrated:

```
cursor = 1000 + 1 = 1001  ->  blockTimestamp_gt: "1001"
rows at ts=1001 included? false
```

The comment sanctions dropping rows "at that exact second" (singular); the code drops two seconds' worth. Fix is `cursor = lastTs;` with `firstPage = true` — `_gt: lastTs` already skips all of `ts == lastTs` and correctly keeps `lastTs + 1`.

Separately: this WARN goes to stdout/markdown only. It sets no `truncated` flag and lands nowhere in the JSON `checks[]`, so a run that dropped rows to saturation can still record `status: "pass"`. Worth a `saturation_events` counter on the returned object, surfaced in the artifact.

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -429,119 +461,318 @@ for (const name of REQUIRED_ENV.slice(1)) {
...
+    if (!requester || ts <= 0) {
+      skippedMissingSenderOrTs += 1;
       continue;
     }
     if (ts > windowEndSec) {
       skippedAfterWindowEnd += 1;
       continue;
     }
+    if (requestId) {
+      withRequestId.push({ requester, ts, title: title ?? null, requestId });
```
OjusWiZard: **Advisory — asymmetric with the analytics side.**

`withRequestId` is pushed to *before* the `if (!title)` continue below, so a title-less row would enter check 2's subgraph feed with `title: null`. The analytics side hard-requires a non-empty `question_title` (`buildAnalyticsRowParityEntry`, L1283), so any such row becomes a false `missing_in_analytics`.

Latent today — the schema declares `questionTitle: String!` and the nested `parsedRequest_` filter excludes null `parsedRequest` — but the code deliberately keeps `skippedMissingTitleOnly` as a "paranoid schema-drift telltale" for exactly the case it then feeds asymmetrically into check 2. Suggest gating the `withRequestId.push` on a non-empty title so the telltale and the check-2 feed agree.

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -429,119 +461,318 @@ for (const name of REQUIRED_ENV.slice(1)) {
...
+  let pages = 0;
   for (;;) {
     const data = await gqlRequest(
       url,
       `query DailyProfitStats {
         dailyProfitStatistics(
           first: ${SUBGRAPH_PAGE}
-          skip: ${skip}
-          where: { date_gte: ${dateGte}, date_lte: ${dateLte} }
+          where: { date_gt: ${cursor}, date_lte: ${dateLte} }
```
OjusWiZard: **Critical — this reintroduces the exact tie-boundary bug the PR fixes for `requests`.**

`date_gt: cursor` is used on **every** page here: there is no `firstPage`/`_gte` branch and no dedup. `fetchMarketplaceRequests` was rewritten to `_gte` + id-dedup precisely because `_gt` loses rows when a full page ends mid-tie (documented at L598-607: "`_gt` reported 6392 rows vs 6393 ... 1 tie-boundary row lost").

`dailyProfitStatistics` is keyed by `(traderAgent, date)`, and this function's own header concedes "many trader agents share the same `date`". Concrete failure: a 1000-row page spans D1..D7 where D7 has 150 rows total but only the first 100 fit. `firstDate !== lastDate`, so the saturation branch does **not** fire; the next page issues `date_gt: D7` and the remaining 50 rows are lost permanently.

Note this is a regression against `main`, which used `date_gte` + `skip` — that lost nothing below the 6000 cap and set `truncated`. Here `truncated` only fires at `pages >= SUBGRAPH_MAX_PAGES` (500), so ordinary multi-page runs lose rows with no counter, no dedup, and no flag.

Impact: `dailyStats.stats` drives `computeSubgraphOpenCount`, so lost settlement rows under-consume the QMR and inflate the subgraph `open` count — a false DIVERGENCE on **check 1, the core gate**. Most exposed on the long `--since/--until` backfills this function's own comment calls out ("the historical consumption horizon can span months").

Fix: mirror `fetchMarketplaceRequests` — `date_gt` on the first page, `date_gte` + dedup on a `(traderAgent.id, date)` composite key thereafter.

scripts/verify-mech-analytics-parity.mjs
OjusWiZard: **Important — off-by-one: this skips an entire extra day.**

Same defect as the `requests` saturation branch, but a day wide: `cursor = lastDate + DAY_SECONDS` feeds `date_gt: cursor`, so every row at `date == lastDate + 86400` — the whole next day of stats — is excluded.

```
lastDate = 100*86400 -> cursor = 8726400 -> date_gt: 8726400
rows at date=8726400 included? false
```

Correct is `cursor = lastDate` (the `date_gt` already excludes `lastDate` itself).

scripts/verify-mech-analytics-parity.mjs
OjusWiZard: **Important — same two defects as the `requests` fetcher, and the header comment overstates parity with it.**

1. `cursor = lastTs + 1` with `blockTimestamp_gt: "${cursor}"` skips all rows at `ts == lastTs + 1` (same off-by-one as L680).
2. The normal path (`cursor = lastTs`) uses `_gt` on every page with no dedup, so this still has the tie-boundary hole that `fetchMarketplaceRequests` was fixed for. `MarketplaceDeliveryWithSignatures` entities can easily share a `blockTimestamp` (multiple batches in one block).

The header at L831-833 claims this is "same shape and truncation / same-timestamp handling as `fetchMarketplaceRequests` above" — it replicates only the saturation branch, not the `_gte`+dedup fix that actually closes the hole.

Blast radius is bounded today since this feed is observability-only, so this reads as Important rather than blocking — but the comment should either be corrected or the fix applied, so nobody later wires this into a gating path trusting it's already correct.

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -682,53 +1117,197 @@ try {
...
+        fetchMarketplaceRequests(
+          marketplaceUrl,
+          windowStartSec,
+          windowEndSec,
+          `${key}:marketplace`
+        ),
+        // Off-chain request IDs from `MarketplaceDeliveryWithSignatures`
+        // entities — the subgraph's `requests(...)` query only returns
+        // on-chain rows (the off-chain handler doesn't create `Request`
+        // entities). Unioned into the subgraph side of check 2 below so
```
OjusWiZard: **Advisory — stale comment, contradicted by the code it sits next to.**

"Unioned into the subgraph side of check 2 below so the two sides model the same superset" was true of `39c06476`, but `f2f7143e` removed the union: L1462 is `subgraphDeliveredParseableRows = marketplace.withRequestId`, no union. The accurate comment is at L818-829 — "OBSERVABILITY-ONLY ... MUST NOT feed row_parity" — which says the opposite.

The code is right; only the comment is wrong. It matters because this is the call site, i.e. the first place someone investigating an off-chain divergence would look, and it would lead them to wrongly rule off-chain rows out as a cause.

scripts/verify-mech-analytics-parity.mjs
OjusWiZard, line 1185: **Critical — this excludes every `source IS NULL` row from both gated checks, and the backfill that would populate `source` has not run.**

Server-side the filter is `AND (%(source)s::text[] IS NULL OR source = ANY(%(source)s::text[]))` — `mech-analytics/etl/api/routes/data.py:358`. With a filter supplied, `NULL = ANY(...)` -> NULL -> the row is **excluded at the API boundary**, never returned. mech-analytics' own docstring says exactly this (`data.py:278-282`, `462-465`): *"Between rollout step 2 (deploy) and step 4 (backfill), a `source` filter would return a near-empty page for the pre-013 tail even against a valid label."*

Migration `013` adds the column nullable; the backfill is a separate one-shot script. mech-analytics#28 merged **2026-08-12 10:08 UTC**, so in production right now most `per_request_scores` rows are still `source IS NULL`.

Two failure modes, and this feed drives **both** checks (`computeAnalyticsOpenCount` at L1319 as well as row_parity at L1463):

**False divergence.** An on-chain request delivered with a parsed title is present on the subgraph side but its analytics row is filtered away -> inflated `missing_in_analytics` -> exit 3.

**False pass.** `analyticsRawRowCount` (L1353) is `scoredRows.rows.length` — the count *after* the server dropped those rows. That is the denominator for both of `resolveCheck1Status`'s guards (zero-ingest, `-lib.mjs:556`; 50% attrition ratio, `-lib.mjs:570-576`). Numerator and denominator shrink together, so the ratio stays healthy however large the NULL tail grows — the guard can only ever measure survivors. Run `--since/--until` over the pre-013 period (documented use case, L126-128): those rows are old and already settled, so they'd be consumed by dailyStats matching and removed from `open` either way, `subgraphOpen === analyticsOpen` holds, and check 1 prints `PASS: open-market counts match exactly` while the dataset it verified is objectively missing rows production ingests.

This also breaks the header invariant at L29-33 — "nothing else is filtered" — since the real consumer applies **no** source filter (`mech-analytics.ts:147-156` passes only `chain_id`/`since`/`limit` + `resolved`|`since_computed_at`).

One more: the PR body says "Depends on nothing to merge (works today)". Not true as of `9147796b` — this needs mech-analytics#28 **deployed**. Against an older deployment FastAPI silently ignores an unknown query param (these are individual `Annotated[..., Query()]` args, no `extra="forbid"`), so the filter no-ops with no error, contradicting the claim at L1002-1004 that an unknown value is "a hard error at the boundary".

Suggested fix: drop the API-side filter and apply `source == null || source in {mech_onchain, ipfs_historical}` locally — that keeps the NULL tail in the comparison until backfill completes and keeps `sourceCounts.null` meaningful. If the filter is kept, `analyticsRawRowCount` needs an unfiltered count (or API total-count metadata) plus a dedicated `analyticsSkippedNullSource` bucket routed into the attrition guard.

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -682,53 +1117,197 @@ try {
...
+        // boundary rather than by incidental drop-out downstream. Off-
+        // chain rows (`mech_offchain`) don't appear in the subgraph's
+        // `Request` collection at all — they're separately fetched via
+        // `fetchMarketplaceOffChainRequestIds` for observability only.
+        // The API is defined by `SourceLabel` in mech-analytics
+        // `etl/api/routes/data.py` (PR #28); typos return 422.
+        fetchScoredRows(chainId, windowStartSec, windowEndSec, `${key}:scored-rows`, {
+          source: ['mech_onchain', 'ipfs_historical'],
+        }),
+        fetchUnscoredRows(chainId, windowStartSec, windowEndSec, `${key}:unscored-rows`),
```
OjusWiZard, line 1187: **Advisory — asymmetric with the scored fetch above.**

`fetchUnscoredRows` gets no `source` filter while `fetchScoredRows` does, so `analytics_unscored_count` includes off-chain rows while `subgraph_off_chain_count` is counted separately — the two observability numbers in the artifact aren't on comparable footings.

Related: neither is gated by any check, so drift between the shell/off-chain populations is now invisible to the exit code. That's a real narrowing of what this script gates versus the previous row_parity definition; worth stating explicitly in the header so operators know those two numbers are eyeball-only.

scripts/verify-mech-analytics-parity.mjs
```diff
@@ -682,53 +1117,197 @@ try {
...
+    // Composition of the scored feed by predict-api ingest source
+    // (mech-analytics PR #28). Live-poller rows are `mech_onchain`;
+    // data-lake backfill is `ipfs_historical`. Rows with `source=null`
+    // are pre-013 tail on the mech-analytics side (the source-backfill
+    // hasn't caught up yet — see mech-analytics
+    // `scripts/backfill_per_request_scores_source.py`). Operators watch
+    // the null count trend to zero as the backfill completes.
+    const sourceCounts = { mech_onchain: 0, ipfs_historical: 0, null: 0 };
+    for (const row of scoredRows.rows) {
+      const label = row.source ?? 'null';
```
OjusWiZard, line 1229: **Important — this counter is now structurally always 0.**

`sourceCounts` iterates `scoredRows.rows`, which is the *already-filtered* result of the `?source=mech_onchain,ipfs_historical` fetch above. A `source IS NULL` row can never appear in that array (`NULL = ANY(...)` is never true), so `sourceCounts.null` reads 0 from the very first run regardless of how much backfill is outstanding.

The comment tells operators to "watch the null count trend to zero as the backfill completes" — the signal added to monitor this risk is disabled by the filter it's meant to monitor, and reads as *backfill done* when it isn't. Either compute it from an unfiltered fetch or drop the counter and the instruction.

OjusWiZard (dismissed):
## Multi-perspective review

Ran the full review stack (code-review, silent-failure, test-coverage) plus manual cross-repo verification against `mech-analytics` (`etl/api/routes/data.py`, migration 013) and the live consumer (`common-util/api/predict/{mech-analytics,roi-distribution}.ts`, `graphql/queries.ts`). Independently confirmed **83/83 tests pass**. I could not verify the "ESLint clean" claim — `yarn install` didn't complete in my sandbox.

The cursor-pagination rewrite, the `_gte`+dedup tie-boundary fix, and the hex/decimal `normaliseRequestId` dedup all look right, and the "displayable rows" rationale is well argued. Findings are inline; two are worth resolving before this gates a cutover.

### Highest-severity, inline
- **`?source=` filter excludes `source IS NULL` rows from both gated checks** (L1177-1179) — and defeats the attrition guard designed to catch that, giving both a false-divergence and a false-pass mode. The backfill that populates `source` hasn't run; mech-analytics#28 merged only today.
- **`fetchDailyStats` reintroduces the `_gt` tie-boundary bug** (L775) that this PR fixes for `requests` — a regression against `main`'s skip-based pagination, silently losing settlement rows on the core gate.

### Not anchorable to a diff line

**Test coverage — the new logic is structurally untestable.** `verify-mech-analytics-parity.mjs` has zero `export` statements, and the 83 passing tests import only from `-lib.mjs`, which this PR changes in comments only. So `normaliseRequestId`, all three pagination loops, the saturation advance, `buildAnalyticsRowParityEntry`, and the analytics dedup are unreachable from the suite. The subprocess tests exercise only the CLI surface (`--help`, missing-env, `--since`/`--until` parsing) and never reach a fetch.

This is worth weighing against the CI job's own rationale in `.github/workflows/main.yml:73-81`: *"two prior wiring regressions (check 1 comparing the wrong bucket, check 3 summing both counters) shipped because these helpers had no CI coverage."* This PR adds +761 lines in exactly that untested surface, and both Critical findings above are the class of bug a unit test would catch.

Minimal refactor: export `normaliseRequestId`, `buildAnalyticsRowParityEntry`, the row-classification block (L706-737), and extract the saturation advance into one shared pure `advanceCursor(page)` — all are pure functions of already-parsed page data and need no new test infrastructure.

**`tsFallbackKeys` is now dead** (L1475-1482). Under the new definition every subgraph row carries a `requestId` and every analytics entry requires a normalised id, so `diffRowSets` never takes the ts path and that warning block can't fire. Either drop it or note it's retained for a future non-id population.

**PR body evidence predates HEAD.** The all-PASS verdict table (Aug 7 -> Aug 11) was produced before `9147796b` added the `?source=` filter — worth a rerun against HEAD before cutover, since that commit is the one carrying the highest-severity finding.

### Checked and clean
- `check2Truncated` excluding off-chain/unscored truncation (L1486-1490) is correct — neither feed is read by `diffRowSets` or any check-2 count. (Separately, neither truncation flag reaches the JSON artifact at all.)
- `withRetry`/`gqlRequest`/`fetchJson` don't degrade failures into empty results; exhausted and non-retryable errors rethrow to the top-level catch and force exit 1. The one soft spot — `body.data ?? {}` at L562 for a 200-OK GraphQL response with `data: null` and no `errors` — yields an empty page, but the vacuous/zero-ingest guards catch that as `gap`/`vacuous`, not `pass`.

OjusWiZard (approved):

atepem (approved):

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Not intentional — went with your second option in `9be137ad`. Dropped `isDelivered: true` and `parsedRequest_ { questionTitle_not: "" }` from the server-side WHERE in `fetchMarketplaceRequests`. Both filters are now applied client-side when building `withRequestId` for check 2, so check 1's `usable` uses the same population the live consumer sees. Check 2's stricter subset comes from the same fetch.

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Applied the full fix in `9be137ad`. `fetchMarketplaceOffChainRequestIds` now uses first-page `blockTimestamp_gt`, subsequent-page `blockTimestamp_gte` + entity id dedup, same shape as `fetchMarketplaceRequests`. Consistency won out over "observability-only lower stakes" since the header claimed parity anyway.

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Not adding a startup probe this round. This script is being run once, post-cutover, against a mech-analytics deployment that we already know is on v0.0.10 (Marc just deployed it, and the source-backfill is running against it right now). A probe would be the right call if this became a scheduled monitor — I'll add one alongside the recurrence in the follow-up refactor, if we ever schedule it.

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Agreed, and I'll do it — but as a follow-up PR rather than in this one. The lift is real (three helpers to extract plus new tests) and I'd rather ship the current fixes now so the cutover isn't blocked on the refactor. Filing it as a follow-up. Once it lands, the two off-by-ones flagged above would have been caught by the extracted `advanceCursor` unit tests, so the CI-coverage rationale is exactly right.

scripts/verify-mech-analytics-parity.mjs
DIvyaNautiyal07: Partially done in `9be137ad`. Trimmed the local blocks in `fetchMarketplaceRequests`, `fetchDailyStats`, `fetchMarketplaceOffChainRequestIds`, and the check-2 section so the two-population split is stated once in the file-level header (check-2 section, L42-L57) and the local sites just describe what the code does. If you spot any remaining near-duplicates, happy to trim further.

Tanya-atatakai (approved):

PR #558, opened 2026-08-18
feat: migrate polymarket data source to the SQD squid

Tanya-atatakai:
## What

Switches the predict-polymarket data source from the graph-node subgraph to the SQD squid (`https://subgraph.autonolas.tech/squid/predict-polymarket/graphql`) and completes the polystrat windowed-ROI cost-basis fix it unblocks.

- All polymarket queries rewritten to the OpenReader dialect (`limit`/`offset`, `orderBy: field_ASC/_DESC`, `squidStatus { height }` instead of `_meta`); two unused queries deleted.
- `profitParticipants` arrives as conditionId strings from the squid — `fetchPolystratDailyStats` resolves titles via a batched `questions(id_in)` lookup before the QMR consumption step (unchanged downstream).
- Windowed polystrat ROI now divides by `dailyTradedSettled` (populated on the squid) instead of the `payout − profit` derivation; the cron writer stores `tradedSettled` on polystrat `byDay` entries.
- Temporary "temporarily shows incorrect values" warnings removed from the polystrat ROI metric and distribution chart.
- Env var renamed `NEXT_PUBLIC_OLAS_POLYMARKET_AGENTS_SUBGRAPH_URL` → `NEXT_PUBLIC_OLAS_POLYMARKET_AGENTS_SQUID_URL`; `/data` methodology snippets updated to the queries the pipeline actually runs.

## Deploy notes

- Define `NEXT_PUBLIC_OLAS_POLYMARKET_AGENTS_SQUID_URL` in Vercel (all environments) before/with this deploy; the old-named var becomes unused.
- Deploy code and env together: old queries fail against the squid and vice versa (crons hold last-good values, so a brief mismatch degrades gracefully).

## Deploy runbook (data cutover)

Existing polystrat `byDay` entries were written without `tradedSettled`, so the ROI blobs must be recreated from scratch under the new code — do this right after deploy:

1. Deploy this PR with `NEXT_PUBLIC_OLAS_POLYMARKET_AGENTS_SQUID_URL` set in Vercel.
2. Delete `metrics-production-20260708-roi-distribution/polystrat-main.json` and `…-requests.json` from the blob store (always as a pair).
3. Call `/api/refresh-metrics/predict-roi-distribution?agent=polystrat` ~7 times sequentially (30 days per run from genesis; done when the blob's `lastDayTimestamp` reaches yesterday).
4. One `/api/refresh-metrics/predict` run; the page catches up within the 5-minute ISR window.

Accuracy and staking-rewards blobs need no action — the accuracy trailing refresh self-heals its recent days from the squid on the first cron run. Rehearsed end-to-end in the dev environment.

## Testing

- `tsc --noEmit` and eslint clean.
- Every rewritten query shape validated against the live squid endpoint, including nested `question.resolution` and the aliased multi-page bets query.
- Squid data spot-checked vs the old subgraph: same trades resolve with full question/resolution linkage (0 of 22,154 bets have a null question, vs a non-zero count on the old deployment).


common-util/api/predict/roi-distribution.ts
```diff
@@ -817,17 +855,14 @@ const computeAgentBlueprintHistogram = (
...
-      //   fields aren't backfilled there yet; revisit when polymarket exposes
-      //   dailyTradedSettled per the same subgraph PR.
+      // 2. Trading costs basis: the per-day settled fields (cost basis of bets
+      //    that resolved that day), summed over the window. Omenstrat entries
+      //    are already 18 decimals; polystrat's are USDC 1e6, scaled below.
+      //    feesSettled is 0 for polystrat (no per-trade fees on Polymarket).
       const tradingCosts = isPolystrat
-        ? scaledPayout - scaledProfit
+        ? totals.tradedSettled * scale
         : totals.tradedSettled + totals.feesSettled;
```
atepem, line 873: **Polystrat `tradedSettled` is missing from every existing `byDay` entry — windowed ROI will be inflated for up to 90 days.**

Before this PR the writer set `tradedSettled` **only** for omenstrat:

```ts
...(agentBlueprint === 'omenstrat' ? { tradedSettled, feesSettled } : {})
```

so every polystrat `byDay[*].agents[*]` already in the `roi-distribution/polystrat-main` blob has `tradedSettled === undefined`. The day cursor (`startDay = existing.lastDayTimestamp + DAY_SECONDS`) never revisits those days, `byDay` keeps 90 days, and `SCHEMA_VERSIONS` was not bumped for `roi-distribution` — so the old blob is reused as-is.

Consequence on the first run after deploy: `if (entry.tradedSettled)` (L838 / L966) skips every legacy day, so an agent that traded across the deploy boundary contributes **profit from all days in the window but cost basis from only the post-deploy days**. One day after deploy the 7D tab divides ~7 days of profit by ~1 day of settled volume. Agents whose whole window predates the deploy get `tradingCosts === 0n` and drop out of the histogram/aggregate entirely. Convergence takes 7/30/90 days per tab.

This ships in the same PR that removes the “temporarily shows incorrect values” warnings, so the wrong numbers are now presented as trustworthy.

Suggested fix: bump the `roi-distribution` entry in `SCHEMA_VERSIONS` and follow the accumulator rollout procedure in CLAUDE.md (copy/rebuild the QMR blobs) so `byDay` is rebuilt from genesis — or backfill `tradedSettled` onto the existing polystrat entries before merging.

common-util/api/predict/roi-distribution.ts
```diff
@@ -207,17 +218,46 @@ const fetchPolystratDailyStats = async (
...
+      return { stats: [], ok: false };
+    }
+  }
+
+  const stats: DailyStatEntry[] = raw.map((stat) => ({
+    ...stat,
+    profitParticipants: (stat.profitParticipants ?? [])
+      .filter((id) => titleById.has(id))
+      .map((id) => ({ metadata: { title: titleById.get(id) as string } })),
+  }));
```
atepem, line 266: **An empty title lookup is silently treated as success and permanently zeroes polystrat mech requests.**

If `titleById` ends up empty — the squid’s `questions.id` is not the conditionId that `profitParticipants` carries, `metadata` is null for freshly indexed markets, or the field is renamed — then `.filter((id) => titleById.has(id))` drops *every* participant, yet the function still returns `ok: true`.

`updateAgentBlueprintData` then writes `mechRequests: 0` for each agent-day, advances `lastDayTimestamp` past them (never recomputed), and leaves the QMR entries open until the `QMR_MAX_AGE_DAYS` TTL flushes them onto their *request* day instead of the settlement day. Net effect: a permanent under-count of polystrat mech costs (i.e. inflated ROI) with no `fetchErrors` entry and nothing in the logs.

Since the id→title contract is now an unverified cross-entity assumption, consider treating “non-empty id set resolved to zero titles” as a fetch failure:

```suggestion
  if (idList.length > 0 && titleById.size === 0) {
    console.error(
      `Polystrat question-title lookup resolved 0 of ${idList.length} conditionIds — treating as a fetch failure`
    );
    return { stats: [], ok: false };
  }

  const stats: DailyStatEntry[] = raw.map((stat) => ({
    ...stat,
    profitParticipants: (stat.profitParticipants ?? [])
      .filter((id) => titleById.has(id))
      .map((id) => ({ metadata: { title: titleById.get(id) as string } })),
  }));
  return { stats, ok: true };
```

common-util/graphql/queries.ts
```diff
@@ -861,7 +871,7 @@ export const getPolymarketTraderAgentsQuery = ({
   skip: number;
 }) => gql`
   query PolymarketTraderAgents {
-    traderAgents(first: ${first}, skip: ${skip}) {
+    traderAgents(limit: ${first}, offset: ${skip}) {
```
atepem: **`traderAgents` lost its implicit ordering — offset paging is now unstable.**

graph-node defaulted `traderAgents(first:, skip:)` to `orderBy: id, orderDirection: asc`, which made the `skip`-based loop in `fetchAllTimeAgents` (`roi-distribution.ts:353`) deterministic. OpenReader emits **no** default `ORDER BY`, so the row order between the `offset: 0` and `offset: 1000` requests is unspecified.

Once polymarket exceeds 1000 trader agents, some agents come back twice (harmless — `agentMap.set` overwrites) and others never at all: they vanish from `allTimeAgents`, so their Max-window ROI disappears from the histogram and from `computeWindowedNetGainAndCosts`, and the `MIN_TRADES_FOR_ROI_DISPLAY` check at L851 stops filtering them in the windowed tabs. The loop still terminates normally via `page.length < LIMIT`, so the failure is silent.

```suggestion
    traderAgents(limit: ${first}, offset: ${skip}, orderBy: id_ASC) {
```

atepem:
**`scripts/verify-mech-analytics-parity.mjs` is broken by the env rename + dialect switch** (outside the diff, so posting here rather than inline).

`scripts/verify-mech-analytics-parity.mjs:216` still declares:

```js
dailyStatsUrlEnv: 'NEXT_PUBLIC_OLAS_POLYMARKET_AGENTS_SUBGRAPH_URL',
```

and that entry is spread into `REQUIRED_ENV` (L225), so after the rename to `..._SQUID_URL` the script aborts with a missing-required-env error before doing any work. `Dockerfile.verify:25` names the old var too.

Even if the old name is supplied, the polystrat path cannot work against the squid:

- `fetchDailyStats` (~L750) sends `dailyProfitStatistics(first: … skip: … orderBy: date orderDirection: asc)` — the graph-node dialect the squid rejects.
- `participantSelection` builds `profitParticipants { metadata { title } }`, but the squid returns plain conditionId strings.
- `participantTitleField: 'metadata.title'` (L219) therefore never matches.

Net: the polystrat half of the migration gate documented in `docs/mech-analytics-migration.md` is silently dead. Worth updating the script in this PR (mirroring the new `getPolymarketDailyProfitStatsQuery` + `getPolymarketQuestionTitlesQuery` two-step), or explicitly noting it as follow-up.

common-util/graphql/queries.ts
```diff
@@ -788,14 +788,15 @@ export const getOmenDailyBrierStatsQuery = ({ date_gte, date_lte, first, skip })
...
     dailyProfitStatistics(
-      first: ${first}
-      skip: ${skip}
-      where: { date_gte: ${date_gte}, date_lte: ${date_lte} }
-      orderBy: date
-      orderDirection: asc
+      limit: ${first}
+      offset: ${skip}
+      where: { date_gte: "${date_gte}", date_lte: "${date_lte}" }
+      orderBy: date_ASC
```
atepem: **Offset paging on a non-unique sort key can drop agent-days permanently.**

`date_ASC` is not unique here — `dailyProfitStatistics` has one row per agent per day, so a single date spans many rows. graph-node appended an implicit `id asc` tiebreaker; OpenReader emits no default `ORDER BY` beyond what's requested, so rows with equal `date` come back in whatever order Postgres returns them, and that order isn't stable across the separate queries that fetch each page.

With `limit`/`offset` paging that means rows straddling a page boundary can be returned twice or skipped entirely. Skips are the damaging case: the day cursor advances once the range is fetched successfully, so a missed agent-day never gets refetched and is absent from `byDay` for good.

Same class as the `traderAgents` finding below. Adding `id_ASC` (or another unique tiebreaker) as a secondary sort makes the page boundaries deterministic.

common-util/graphql/queries.ts
```diff
@@ -1178,10 +1104,9 @@ export const getPolymarketBetsWithBettorQuery = ({
   for (let i = 0; i < pages; i++) {
     queries.push(`
       page${i}: bets(
-        first: ${first}
-        skip: ${i * first}
-        orderBy: blockTimestamp
-        orderDirection: desc
+        limit: ${first}
+        offset: ${i * first}
+        orderBy: blockTimestamp_DESC
```
atepem: **Ten offset pages ordered only by `blockTimestamp_DESC` — ties are the norm, not the exception.**

Every bet mined in the same block shares a `blockTimestamp`, and on Polygon that's a lot of rows per value. Under graph-node the implicit `id asc` tiebreaker made the total order deterministic, so `offset: i * first` sliced a stable sequence. OpenReader adds no such tiebreaker.

These ten aliased pages resolve as ten independent SQL queries, so each is free to order its tied rows differently. Bets near a page boundary can then appear in two pages or in none, which feeds double-counted or missing bets straight into the tool-accuracy won/total ratios.

Adding `id_DESC` as a secondary sort restores the stable total order the offsets assume.

common-util/api/predict/accuracy.ts
```diff
@@ -150,8 +151,7 @@ const fetchPolyDayBuckets: FetchDayBuckets = async (
     )) as PolyBetsResponse;
 
     if (!metaChecked) {
-      if (response?._meta?.hasIndexingErrors) indexingErrors.push('predict:polygon');
-      if (chainBlock && checkSubgraphLag(chainBlock, response?._meta?.block?.number, 'polygon')) {
+      if (chainBlock && checkSubgraphLag(chainBlock, response?.squidStatus?.height, 'polygon')) {
```
atepem: **The polymarket accuracy path now has no health signal that can actually fire.**

The `_meta { hasIndexingErrors }` check is gone from this query, and `indexingErrors` is only ever pushed to from the gnosis path (line 90) — so for polystrat, lag is the sole remaining signal.

But `checkSubgraphLag` returns `false` when the block number is `undefined` ([metric-utils.ts:96](https://github.com/valory-xyz/olas-website/blob/main/common-util/graphql/metric-utils.ts#L96)):

```ts
if (!chainBlock || subgraphBlockNumber === undefined) return false;
```

If the squid omits `squidStatus` or returns a null `height` — exactly what a stalled or mid-restart squid tends to do — the lag check reports healthy, `createStaleStatus` sees empty arrays, and a stale accuracy number publishes with no `stale` flag and no grey-out.

Worth treating a missing `squidStatus.height` as a `fetchError` rather than as "no lag", so a status-less squid degrades visibly instead of silently.

atepem:
**Nit — stale comment on `Totals`** ([roi-distribution.ts:817](https://github.com/valory-xyz/olas-website/blob/feat/polymarket-squid-migration/common-util/api/predict/roi-distribution.ts#L817))

```ts
tradedSettled: bigint; // omenstrat only — 0n for polystrat
```

Seven lines below, the windowed cost basis now divides by `tradedSettled` for polystrat too, so the comment asserts the opposite of what the code does. Since the field's polystrat behaviour is the crux of the migration comment above, this one is worth correcting rather than leaving to drift.

common-util/api/predict/roi-distribution.ts
Tanya-atatakai: Resolved operationally rather than in code: the deploy procedure recreates the polystrat ROI blobs completely from scratch — `roi-distribution/polystrat-main` + `-requests` are deleted after the new code is live, then the distribution cron is re-run until it reaches head, so every retained `byDay` entry is written by the new code with `tradedSettled` populated. No legacy entries survive the cutover (rehearsed end-to-end in the dev environment). The runbook is now part of the PR description. We deliberately avoided the `SCHEMA_VERSIONS` bump because the version key covers the whole `roi-distribution` family and would also reset the omenstrat blobs.

common-util/api/predict/roi-distribution.ts
Tanya-atatakai: Applied in f6d988b4 — a lookup that resolves 0 of N conditionIds now returns `ok: false` so the day cursor doesn't advance past the chunk.

common-util/graphql/queries.ts
Tanya-atatakai: Fixed in f6d988b4 (`orderBy: id_ASC`).

common-util/graphql/queries.ts
Tanya-atatakai: Fixed in f6d988b4 (`orderBy: [date_ASC, id_ASC]`, multi-key orderBy verified against the live squid).

common-util/graphql/queries.ts
Tanya-atatakai: Fixed in f6d988b4 (`orderBy: [blockTimestamp_DESC, id_DESC]`).

common-util/api/predict/accuracy.ts
Tanya-atatakai: Fixed in f6d988b4 — a missing `squidStatus.height` now marks `predict:polygon` as lagging (stale grey-out) instead of passing as healthy.

Tanya-atatakai:
Re `scripts/verify-mech-analytics-parity.mjs`: agreed it's affected by the rename + dialect switch, but the verify scripts are deliberately out of scope for this PR — it doesn't touch them (the diff is the 11 migration files only). Updating (or retiring) the parity gate for the squid dialect will be handled as a separate follow-up.

Re the stale `Totals` comment: fixed in f6d988b4.

All other inline findings are addressed in f6d988b4 — see the per-thread replies.

atepem (approved):

ROMAN-VALORY (approved):

PR #559, opened 2026-08-24
feat: add Polystrat to the Agent Economy Explorer

Tanya-atatakai:
Closes BDMARK-2261.

Adds Polystrat as a second agent under the Predict economy in the Agent Economy Explorer, using the SQD squid (introduced in #558) as the accuracy data source.

## What's included

- **Agent sub-toggle for Predict** — same switcher Babydegen uses; Omenstrat (purple) / Polystrat (indigo). Deep-link `?economy=predict` still lands on Omenstrat first.
- **Data** (`common-util/api/explorer.ts`):
  - DAA + transactions from the Polygon registry subgraph, agentId 86 (`POLYSTRAT_AGENT_CLASSIFICATION.valory_trader`). `fetchRegistrySeries` is parameterized by chain + agent IDs so Omenstrat/Polystrat share the paging/lag/transform code.
  - Daily prediction accuracy from the predict-polymarket squid (`getPolymarketBetsByTimeRangeQuery`, cursor-paged newest-backward, merged into the stored series like Omenstrat's). Unresolved bets are skipped and picked up once resolved.
  - Per-source sparsity threshold: Polystrat omits days with <3 resolved bets (vs Omenstrat's 10) — Polystrat places ~2–15 bets/day, so 10 would blank most days.
  - New `polystrat` key in the explorer snapshot; cron threads `polystratPrevious` and reports `polystratCounts`.
- **Indigo heatmap ramp** anchored on the Polystrat icon's deep navy (`#14127a`); light levels route through mid indigo so they stay distinguishable from the empty-cell grey (same shape as the teal ramp). Clearly darker than Basius's Base blue.
- **Launch marker** ringed on 2026-02-10 ("Polystrat launched publicly"), Basius-style.
- **Accuracy tile tooltip** ("Average daily win rate, based on bets placed each day…") via the same info popover the DAA tile uses; the tile popover now wraps long text (`max-w-[320px]` instead of `whitespace-nowrap`).

## Rollout

Nothing manual needed: the daily 06:00 UTC explorer cron populates the new snapshot key, and Polystrat's full history (~3–4k bets since late January) fits inside the default 12-page accuracy window, so the first run backfills to genesis. Hit `/api/refresh-metrics/explorer` once manually to populate sooner. Page is ISR with 1h revalidate.

## Testing

- `tsc --noEmit` + ESLint clean.
- Verified locally against the dev-scoped snapshot: sub-toggle, indigo ramp + legend, Feb 10 marker, and all three tiles populated; bet-volume distribution checked directly against the squid.

## Screenshots

image

### DIFF
--- a/common-util/api/explorer.ts
+++ b/common-util/api/explorer.ts
@@ -25,6 +25,8 @@ import {
   getExplorerDailyProfitStatsQuery,
   getPolymarketBetsByTimeRangeQuery,
   mechAtaTransactionsQuery,
+  PolymarketBetRow,
+  PolymarketBetsResponse,
 } from 'common-util/graphql/queries';
 import { MetricWithStatus, WithMeta } from 'common-util/graphql/types';
 import { getMidnightUtcTimestampDaysAgo } from 'common-util/time';
@@ -155,8 +157,8 @@ const transformExplorerSeries = (
 
 // Page a registry subgraph's daily per-agent rows (the subgraph caps `first` at 1000,
 // so we skip-page until a short page). Chain-agnostic — `explorerOmenstratSeriesQuery`
-// is parameterised by `agentIds`, so the same query serves Omenstrat (Gnosis) and
-// Babydegen (Optimism + Mode). Returns the last page's `_meta` for lag/error detection.
+// is parameterised by `agentIds`, so the same query serves every registry-backed fleet
+// across chains. Returns the last page's `_meta` for lag/error detection.
 const pageRegistryRows = async (
   client: GraphQLClient,
   agentIds: number[],
@@ -473,7 +475,9 @@ export const fetchOmenstratExplorerSeries = async (
   return {
     value: {
       ...registry.value,
-      // Window-merge into prior history: a failed fetch ([]) leaves `previous` intact.
+      // Window-merge into prior history: fresh days overwrite stored ones. A fully-failed
+      // fetch ([]) leaves `previous` intact, but a partially-failed one can overwrite the
+      // oldest reached day with a partial sample (healed on the next full-window run).
       accuracy: mergeSeries(previous?.accuracy, accuracy ?? []),
       roi: mergeSeries(previous?.roi, roi ?? []),
     },
@@ -495,25 +499,24 @@ export type PolystratExplorerSeries = RegistrySeries & {
   accuracy: DaaSeriesPoint[];
 };
 
-type PolymarketBet = {
-  blockTimestamp: string;
-  outcomeIndex: number;
-  question: { resolution: { winningIndex: number } | null } | null;
-};
-
 /**
  * Daily Prediction Accuracy for Polystrat — same shape and cursor paging as
  * fetchOmenstratDailyAccuracy, but sourced from the predict-polymarket squid. The
  * squid has no server-side resolution filter, so unresolved bets come back and are
  * skipped in code — they get counted on a later run, once their market resolves and
- * the recent window is re-fetched.
+ * the recent window is re-fetched. Also returns the squid's indexed height (null if
+ * no page succeeded) so the caller can lag-check it — a stalled squid returns a
+ * successful-looking response with frozen data, so height is the only freshness signal.
  */
-const fetchPolystratDailyAccuracy = async (maxPages: number): Promise<DaaSeriesPoint[]> => {
-  const bets: PolymarketBet[] = [];
+const fetchPolystratDailyAccuracy = async (
+  maxPages: number
+): Promise<{ series: DaaSeriesPoint[]; squidHeight: number | null }> => {
+  const bets: PolymarketBetRow[] = [];
+  let squidHeight: number | null = null;
   const pages = Math.min(maxPages, ACCURACY_MAX_PAGES);
   let cursor = ACCURACY_CURSOR_START;
   for (let page = 0; page < pages; page += 1) {
-    let pageRows: PolymarketBet[];
+    let pageRows: PolymarketBetRow[];
     try {
       const data = (await requestWithRetry(() =>
         polymarketAgentsGraphClient.request(
@@ -523,8 +526,9 @@ const fetchPolystratDailyAccuracy = async (maxPages: number): Promise<DaaSeriesP
             blockTimestamp_lt: cursor,
           })
         )
-      )) as { bets: PolymarketBet[] };
+      )) as PolymarketBetsResponse;
       pageRows = data?.bets ?? [];
+      squidHeight = squidHeight ?? data?.squidStatus?.height ?? null;
     } catch (error) {
       console.error(
         `${POLYSTRAT_ACCURACY_SOURCE}: page ${page} failed after retries; keeping ${bets.length} bets`,
@@ -551,10 +555,11 @@ const fetchPolystratDailyAccuracy = async (maxPages: number): Promise<DaaSeriesP
     byDay.set(date, entry);
   });
 
-  return Array.from(byDay.entries())
+  const series = Array.from(byDay.entries())
     .filter(([, e]) => e.total >= POLYSTRAT_MIN_BETS_PER_DAY)
     .map(([date, e]) => ({ date, count: Math.round((e.correct / e.total) * 100) }))
     .sort((a, b) => a.date.localeCompare(b.date));
+  return { series, squidHeight };
 };
 
 /**
@@ -570,26 +575,43 @@ export const fetchPolystratExplorerSeries = async (
   const timestampLt = getMidnightUtcTimestampDaysAgo(0);
   const timestampGt = getMidnightUtcTimestampDaysAgo(SERIES_WINDOW_DAYS);
 
-  const [registry, accuracy] = await Promise.all([
+  // fetchPolystratDailyAccuracy swallows per-page errors internally (a total failure
+  // surfaces as squidHeight null), so no .catch is needed here.
+  const [registry, accuracy, chainBlock] = await Promise.all([
     fetchRegistrySeries('polygon', POLYSTRAT_AGENT_IDS, timestampGt, timestampLt),
-    fetchPolystratDailyAccuracy(accuracyPages).catch((error) => {
-      console.error(`Error fetching ${POLYSTRAT_ACCURACY_SOURCE}:`, error);
-      return null;
-    }),
+    fetchPolystratDailyAccuracy(accuracyPages),
+    getChainBlockNumber('polygon'),
   ]);
 
   if (!registry.value) {
     return { value: null, status: registry.status };
   }
 
+  // The squid doesn't error when unhealthy — it stops advancing — so a missing or
+  // lagging height marks the accuracy source stale (indicator only; registry data
+  // stays live, so nothing is frozen).
+  const accuracyLagging =
+    accuracy.squidHeight == null || checkSubgraphLag(chainBlock, accuracy.squidHeight, 'polygon');
+  const status = accuracyLagging
+    ? createStaleStatus({
+        indexingErrors: registry.status.indexingErrors,
+        fetchErrors: registry.status.fetchErrors,
+        laggingSubgraphs: [
+          ...(registry.status.laggingSubgraphs ?? []),
+          POLYSTRAT_ACCURACY_SOURCE,
+        ],
+      })
+    : registry.status;
+
   return {
     value: {
       ...registry.value,
-      // Window-merge into prior history: a failed fetch ([]) leaves `previous` intact.
-      accuracy: mergeSeries(polystratPrevious?.accuracy, accuracy ?? []),
+      // Window-merge into prior history: fresh days overwrite stored ones. A fully-failed
+      // fetch ([]) leaves `previous` intact, but a partially-failed one can overwrite the
+      // oldest reached day with a partial sample (healed on the next full-window run).
+      accuracy: mergeSeries(polystratPrevious?.accuracy, accuracy.series),
     },
-    // Registry drives staleness; accuracy is best-effort.
-    status: registry.status,
+    status,
   };
 };
 
--- a/common-util/api/explorer.ts
+++ b/common-util/api/explorer.ts
@@ -1,10 +1,14 @@
 import { GraphQLClient } from 'graphql-request';
 
-import { OMENSTRAT_AGENT_CLASSIFICATION } from 'common-util/constants';
+import {
+  OMENSTRAT_AGENT_CLASSIFICATION,
+  POLYSTRAT_AGENT_CLASSIFICATION,
+} from 'common-util/constants';
 import { DaaSeriesPoint } from 'common-util/explorer';
 import {
   BABYDEGEN_GRAPH_CLIENTS,
   MARKETPLACE_GRAPH_CLIENTS,
+  polymarketAgentsGraphClient,
   predictAgentsGraphClient,
   REGISTRY_GRAPH_CLIENTS,
 } from 'common-util/graphql/client';
@@ -19,6 +23,7 @@ import {
   explorerOmenstratSeriesQuery,
   getExplorerBetsQuery,
   getExplorerDailyProfitStatsQuery,
+  getPolymarketBetsByTimeRangeQuery,
   mechAtaTransactionsQuery,
 } from 'common-util/graphql/queries';
 import { MetricWithStatus, WithMeta } from 'common-util/graphql/types';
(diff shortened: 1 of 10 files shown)


common-util/api/explorer.ts
```diff
@@ -471,6 +482,117 @@ export const fetchOmenstratExplorerSeries = async (
...
+    try {
+      const data = (await requestWithRetry(() =>
+        polymarketAgentsGraphClient.request(
+          getPolymarketBetsByTimeRangeQuery({
+            first: ACCURACY_BET_PAGE_SIZE,
+            blockTimestamp_gte: 0,
+            blockTimestamp_lt: cursor,
+          })
+        )
+      )) as { bets: PolymarketBet[] };
```
OjusWiZard: **[Important] The squid's freshness signal is fetched over the wire, then discarded.**

`getPolymarketBetsByTimeRangeQuery` always selects `squidStatus { height }`, but this cast takes only `{ bets }` and never reads it. `fetchPolystratExplorerSeries` then returns `status: registry.status` (the *Polygon registry* status), so nothing about squid health reaches the published metric.

I initially wondered whether this was just the pre-existing Omenstrat pattern, but I don't think it is: `getExplorerBetsQuery` selects no `_meta` at all, so there's no freshness signal there to drop. The squid one does return it, and `common-util/api/predict/accuracy.ts:150-155` already consumes it correctly — with the comment *"The squid handles errors differently (it stops advancing), so the lag check covers them."*

That's the crux: a stalled squid returns a **successful-looking** response with frozen data. A subgraph errors; a squid goes quiet. So the failure mode is: squid stalls, Polygon registry keeps indexing fine, `status.stale` stays `false`, and the accuracy tile freezes at its last value with no "data may be delayed" banner.

Fix mirrors `accuracy.ts` closely — read `response.squidStatus?.height`, run `checkSubgraphLag(chainBlock, height, 'polygon')`, and union the result into the returned status.

common-util/api/explorer.ts
```diff
@@ -471,6 +482,117 @@ export const fetchOmenstratExplorerSeries = async (
...
+  }
+
+  return {
+    value: {
+      ...registry.value,
+      // Window-merge into prior history: a failed fetch ([]) leaves `previous` intact.
+      accuracy: mergeSeries(polystratPrevious?.accuracy, accuracy ?? []),
+    },
+    // Registry drives staleness; accuracy is best-effort.
+    status: registry.status,
```
OjusWiZard: Follow-on from the `squidStatus` comment above: because status is wired entirely to the Polygon registry, no accuracy-source outcome — stalled squid, failed page, empty result — can ever set `status.stale`.

`// Registry drives staleness; accuracy is best-effort` accurately describes the code, but "best-effort" here means a frozen accuracy series renders as live and healthy. Unioning the squid's lag signal in would make the existing UI banner fire without any UI change.

common-util/api/explorer.ts
```diff
@@ -471,6 +482,117 @@ export const fetchOmenstratExplorerSeries = async (
...
+  ]);
+
+  if (!registry.value) {
+    return { value: null, status: registry.status };
+  }
+
+  return {
+    value: {
+      ...registry.value,
+      // Window-merge into prior history: a failed fetch ([]) leaves `previous` intact.
```
OjusWiZard: **[Important] This comment doesn't match the code.**

*"a failed fetch (`[]`) leaves `previous` intact"* is only true for a **total** failure. `fetchPolystratDailyAccuracy` catches per-page errors and `break`s, returning a **partial** array — which `mergeSeries` then writes over `previous` for the dates it covers (`byDate.set` unconditionally, no completeness check).

Concretely: a mid-loop page error can leave the oldest in-flight day counted from a fragment of its real bets. `POLYSTRAT_MIN_BETS_PER_DAY = 3` makes a fragment likelier to clear the publish bar than Omenstrat's `10` does.

Worth saying this is **self-limiting** rather than serious: Polystrat's full history fits inside the default window, so every run re-fetches from genesis and heals the previous run's damage. It stops being self-healing if volume ever outgrows the 12-page window.

Suggested wording: *"fresh days overwrite stored ones; a fully-failed fetch leaves `previous` intact, but a partially-failed one can overwrite the oldest reached day with a partial sample."*

Also note the outer `.catch((error) => null)` on this call is effectively dead — the inner loop swallows every error, so it can't fire.

common-util/api/explorer.ts
```diff
@@ -471,6 +482,117 @@ export const fetchOmenstratExplorerSeries = async (
   };
 };
 
+// ── Polystrat (Polygon registry + predict-polymarket squid) ──────────────────
+const POLYSTRAT_ACCURACY_SOURCE = 'predict:polygon:explorer-accuracy';
+// Lower than Omenstrat's MIN_BETS_PER_DAY: Polystrat places ~2-15 bets/day (vs hundreds
+// on Omen), so 10 would blank most days. 3 still drops the only-extremes days (1-2 bets
+// can only show 0/50/100%).
+const POLYSTRAT_MIN_BETS_PER_DAY = 3;
```
OjusWiZard: The rationale for `3` is sound for **heatmap density** — at ~2–15 bets/day, `10` really would blank most cells.

Flagging that this constant is doing double duty though: it's also the sample floor for the headline tile, which averages these daily percentages (`components/ExplorerPage/index.tsx:114`). A 3-bet day can only read 0/33/67/100%, and it counts as much as a 15-bet day. See the inline note there — if the headline moves to bet-weighted, this threshold can stay at `3` purely for the heatmap, which is probably the cleanest split.

common-util/api/explorer.ts
```diff
@@ -471,6 +482,117 @@ export const fetchOmenstratExplorerSeries = async (
...
+
+/** Polystrat daily series — registry DAA/transactions + squid-derived accuracy. */
+export type PolystratExplorerSeries = RegistrySeries & {
+  /** Prediction Accuracy — daily win-rate (%) of resolved bets; days below POLYSTRAT_MIN_BETS_PER_DAY omitted. */
+  accuracy: DaaSeriesPoint[];
+};
+
+type PolymarketBet = {
```
OjusWiZard: This type is byte-identical to `PolyBetRow` in `common-util/api/predict/accuracy.ts:123-127` — same squid, same query, and both applied via an unchecked `as` cast.

Two independently-published accuracy numbers now depend on it (this heatmap and the Predict page tile). If the squid schema shifts, nothing structurally forces both files to be touched, and since neither validates the response the failure mode is a wrong percentage rather than a crash.

Cheap fix: hoist one exported `PolymarketBetRow` next to the query in `queries.ts` and import it in both places.

common-util/api/explorer.ts
```diff
@@ -471,6 +482,117 @@ export const fetchOmenstratExplorerSeries = async (
...
+};
+
+/**
+ * Daily Prediction Accuracy for Polystrat — same shape and cursor paging as
+ * fetchOmenstratDailyAccuracy, but sourced from the predict-polymarket squid. The
+ * squid has no server-side resolution filter, so unresolved bets come back and are
+ * skipped in code — they get counted on a later run, once their market resolves and
+ * the recent window is re-fetched.
+ */
+const fetchPolystratDailyAccuracy = async (maxPages: number): Promise<DaaSeriesPoint[]> => {
```
OjusWiZard: This shares ~75-80% of its body with `fetchOmenstratDailyAccuracy` above — identical page loop, cursor advance, break conditions, `byDay` reduce, and `filter → map → sort`. Only the client/query, the resolved-and-won predicate, and the threshold differ.

Calling it out mainly because sharing was this PR's own stated goal, and it landed for `fetchRegistrySeries` — the accuracy path just got hand-copied instead. A descriptor like `{ client, query, getTimestamp, outcome, minBetsPerDay, source }` behind one shared paging body would take the combined ~90 lines to ~65 and, more usefully, make the cursor tie-break and paging-cap logic single-source.

Not a blocker, and reasonable to defer — but the duplication is already three-way if you count `fetchPolyDayBuckets` in `predict/accuracy.ts`, which pages the same query with the same win predicate.

components/ExplorerPage/index.tsx
```diff
@@ -111,6 +112,8 @@ const METRIC_CONFIG: Record<string, MetricDef> = {
     // Unweighted mean of the daily win-rates we have data for.
     headline: (s) =>
       s.length ? `${Math.round(s.reduce((sum, p) => sum + p.count, 0) / s.length)}%` : '--',
```
OjusWiZard: **[Important] This headline won't agree with the Predict page's number for the same agent.**

This is an **unweighted mean of daily win-rates** (`sum(daily %) / day count`), while `buildWindowedAccuracy` in `common-util/api/predict/accuracy.ts` sums `won/total` across the window — **weighted**.

For Omenstrat (hundreds of bets/day) the two nearly converge, so this hasn't mattered. For Polystrat at ~2–15 bets/day with a floor of 3 they won't: a 3-bet day can only read 0/33/67/100%, and it carries the same weight as a 15-bet day. Net effect is two different "Polystrat accuracy" figures on the same site with nothing reconciling them.

The new tooltip does disclose the method, which genuinely mitigates it. But bet-weighting the headline would be a small change and would make the two pages agree — or raise `POLYSTRAT_MIN_BETS_PER_DAY`, though that trades away heatmap density (see the note on `explorer.ts:490`).

components/ExplorerPage/index.tsx
```diff
@@ -149,6 +153,13 @@ const ECONOMY_META: Record<string, EconomyMeta> = {
         icon: '/images/predict-page/omenstrat-icon.png',
         ramp: 'purple',
       },
+      {
+        key: 'polystrat',
+        label: 'Polystrat',
+        icon: '/images/predict-page/polystrat-icon.png',
+        ramp: 'indigo',
+        marker: { date: '2026-02-10', label: 'Polystrat launched publicly' },
```
OjusWiZard: Worth double-checking this date: `POLYMARKET_GENESIS_TS` in both `predict/roi-distribution.ts:39` and `predict/staking-rewards.ts:21` is `1768867200` = **2026-01-20**, three weeks before this marker.

So the heatmap will show ~3 weeks of activity *before* its own "launched publicly" ring, which reads oddly. Entirely plausible if a soft launch preceded the public one — just confirming that's intended and not an off-by-a-few-weeks.

pages/agent-economies/explorer.tsx
```diff
@@ -30,6 +30,10 @@ export const getStaticProps = async () => {
         series: data?.omenstrat?.value ?? { daa: [], transactions: [], accuracy: [], roi: [] },
         status: data?.omenstrat?.status ?? null,
       },
+      polystrat: {
+        series: data?.polystrat?.value ?? { daa: [], transactions: [], accuracy: [] },
```
OjusWiZard: **[Important] This fallback renders a real-looking `0` rather than a no-data state.**

`daa.headline` is `formatCount(s.length ? ... : 0)` → `"0"` on an empty series, and `selectable: () => true` keeps the tile undimmed (`components/ExplorerPage/index.tsx:82`); `transactions` reduces to `0` the same way (`:94`). `accuracy` and `aum` correctly return `'--'`.

So between deploy and the first `refresh-metrics/explorer` cron run, Predict → Polystrat shows **"Latest DAAs: 0"** and **"Total Transactions: 0"** as ordinary black numbers — indistinguishable from genuinely zero agent activity. The same path fires again later if a run's Polygon registry fetch fails with no prior Polystrat entry to fall back on.

Omenstrat and Babydegen have years of backfill so their empty branch is effectively dead code; Polystrat's is near-guaranteed to be hit. The "hit the endpoint manually" note in the PR description narrows the window but doesn't close it.

Suggest `'--'` + `selectable: (s) => s.length > 0` for `daa`/`transactions`, matching what `accuracy`/`aum` already do.

pages/api/refresh-metrics/explorer.ts
```diff
@@ -64,6 +67,11 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
         accuracy: value.accuracy.length,
         roi: value.roi.length,
       },
+      polystratCounts: polystrat && {
```
OjusWiZard: Minor: `seriesCounts` (just above, line 54) already generalizes this exact `{daa, transactions, +optional third}` shape and is used for the three Babydegen agents — this block and the `counts` block above it both hand-roll it instead.

Widening `seriesCounts` to report lengths for whatever array-valued keys are present would let `counts`, `polystratCounts` and the Babydegen entries share one helper (`mech` legitimately stays separate for its derived `ataTotal`). Debug-only JSON, so zero risk — purely a tidy-up.

Also relevant here: the budget comment at line 19 (*"≈ accuracyPages 400"*) now describes a knob that drives two accuracy loops. Details in the review body — short version is it's a comment fix, not a real risk at current volume.

OjusWiZard (dismissed):
## Multi-perspective review — approve with changes

Ran a full review pass (correctness, comments, tests, silent failures, type design, simplification). **No Critical findings.** The change follows the Basius sub-tab precedent (`a6d6118`) file-for-file, and I confirmed `tsc --noEmit` and ESLint are clean on all six changed files.

Approving — the items below are worth a follow-up but none block. Two of them ship user-visible wrong-looking numbers, so I'd suggest picking those up before or shortly after merge.

### Important
1. **The squid's freshness signal is fetched and discarded** — `common-util/api/explorer.ts:526`. See inline.
2. **Pre-first-cron Polystrat tab renders a real-looking `0`** — see inline on `pages/agent-economies/explorer.tsx:34`.
3. **Explorer accuracy won't match the Predict page's for the same agent** — see inline on `components/ExplorerPage/index.tsx:114`.
4. **Stale comments** — see inline, plus the two out-of-diff ones below.

### Findings outside the diff (can't be anchored inline)

- **`common-util/api/explorer.ts:157`** — the comment above `pageRegistryRows` says it serves *"Omenstrat (Gnosis) and Babydegen (Optimism + Mode)"*. After this PR it also serves Polystrat (Polygon), Babydegen/Base, and Mech. Worth re-wording to something that won't need editing per agent, e.g. *"parameterised by `agentIds`, serving Omenstrat, Polystrat, Babydegen and Mech across chains."*

- **`pages/api/refresh-metrics/explorer.ts:19`** — `// Keep params under what fits the 300s function limit (≈ accuracyPages 400 / roiDays 365)` predates this PR adding a second accuracy fetch on the same knob. Worth a note that `accuracyPages` now drives two loops. To be clear though, I'd **not** treat this as a real operational risk today: the loop breaks on the first short page, and Polystrat exhausts its ~3–4k-bet history in ~4 pages, so `accuracyPages=400` yields ~4 squid requests, not 400. It only starts to matter if Polystrat exceeds ~12k bets.

- **`components/ExplorerPage/index.tsx:82` and `:94`** — `daa.headline` returns `formatCount(0)` → `"0"` on an empty series and `selectable: () => true` keeps the tile undimmed; same for `transactions`. Compare `accuracy` (`:114`) and `aum` (`:124`), which correctly return `'--'`. This is the config behind the inline finding on `explorer.tsx:34`.

- **`ECONOMY_META.metrics: string[]`** is unchecked against `METRIC_CONFIG`. A typo (`'accuray'`) compiles and then crashes the page at `config.tileLabel`. `type MetricKey = keyof typeof METRIC_CONFIG` closes it.

- **Agent switch deliberately keeps the active metric** (`index.tsx:322-327`). Newly reachable for Predict now that it has a sub-toggle: switching Omenstrat → Polystrat while on `accuracy` with an empty Polystrat series leaves `accuracy` active-but-dimmed over a blank heatmap.

### Verified clean — checked because they looked like likely problems

- **No `SCHEMA_VERSIONS` bump needed.** Adding `polystrat` is purely additive; `mergeSnapshotTree` carries missing keys from either side, and there's a passing required-CI test for exactly this shape (`metric-status.test.mjs:251`, *"a metric gaining a new sibling keeps both"*). The Basius PR added a snapshot key the same way without a bump.
- **Polygon wiring is complete** — `REGISTRY_GRAPH_CLIENTS.polygon`, `CHAIN_CONFIG.polygon` (48h tolerance) and `POLYGON_RPC` all exist, so lag detection genuinely works on the new chain.
- **`POLYSTRAT_AGENT_IDS` sharing the constant's array by reference is safe** — read-only through every consumer. The adjacent Omenstrat comment is about *values* (needing legacy id 12), not mutation safety.
- **`HeatmapRamp` / `HEATMAP_RAMPS`** is compiler-enforced in both directions, and the explicit ramp tables are the right call over a generator — the stored hex values aren't an exact function of two endpoints, so a formula would risk a subtle visual change.
- **Shipping without tests is consistent with repo convention.** CI has three `node --test` suites, two required, but none cover `common-util/api/*`, and comparable prior PRs (including a larger ROI-aggregation rewrite) also shipped without tests. If one cheap test is ever wanted here, extracting the shared day-bucket-and-threshold reduce and testing the `>= 3` boundary would cover Omenstrat and Polystrat at once.

<sub>Reviewed with Claude Code.</sub>

ROMAN-VALORY (dismissed):

components/ExplorerPage/index.tsx
Tanya-atatakai: Keeping the unweighted daily mean here. Two reasons: the stored series only carries one percentage per day (the won/total pair is discarded at fetch time), so bet-weighting would need a snapshot-shape extension; and each tile now fully discloses its own method — the Explorer tooltip explicitly says every day is weighted equally regardless of bet count (b40b0b7c). Happy to revisit weighting if the two figures diverge confusingly in practice.

components/ExplorerPage/index.tsx
Tanya-atatakai: Good catch — both dates are real: 2026-02-10 is the official public launch (the marker), but we were testing internally before that, with first on-chain activity on 2026-01-16. Updated POLYMARKET_GENESIS_TS/_DAY to 1768521600 (2026-01-16) in b40b0b7c so genesis matches first activity; the marker stays on the public launch.

ROMAN-VALORY (approved):

OjusWiZard (approved):

PR #36, opened 2024-02-25
Fix up and Add new images and components for LearnPage

oaksprout:
https://olas-website-ishi4iw8g-autonolas.vercel.app/learn

- Re-add 'what's an autonomous service" image
image

- Fix index of education articles
- image

- Adds pages for education articles
- image

- Add Learn to Resources menu
image

- Adds What could you build links
image

- deletes mission
- replaces all Autonolas with Olas
-

PR #560, opened 2026-08-27
feat: flip USE_MECH_ANALYTICS default to on

DIvyaNautiyal07:
## Summary

Flips the `USE_MECH_ANALYTICS` flag default from opt-in (`=== 'true'`) to opt-out (`!== 'false'`) so mech-analytics is the primary source for the per-request mech feed and `senderTotal` on the predict page. Rollback to the subgraph is a one-liner env-var flip.

## Prerequisites (all landed)

- **mech-analytics PR #36** (valory-xyz/mech-analytics#36) — undelivered on-chain requests now land as shell rows so `COUNT(*)` matches on-chain reality. Deployed and verified live: `~292k` historical undelivered rows backfilled, live scoring flowing.
- **Placeholder repair** — one-shot rewrote 204 historical trader-tool rows that had `prompt='[offchain request]'` (mech-side bug that shipped before valory-xyz/mech#478). Real question text pulled back from the marketplace subgraph.
- **Parity re-run** (Aug 1-25 window) after both fixes landed:
  - `row_parity.missing_in_analytics`: **225 → 21** (matches the 204 repair count)
  - `senderTotal`: 302,465 = 302,465 (PASS, delta 0 on omenstrat, delta 1 on polystrat within ±5 tolerance)
  - Residual 21 missing are 2 mech-side attribution edge cases + ~19 non-trader Safes using trader tools; none affect user-visible metrics

## Test plan

- [ ] Deploy to staging, `USE_MECH_ANALYTICS` unset → reads should flow through mech-analytics; verify ROI + senderTotal render correctly on `/predict`
- [ ] Verify rollback path: set `USE_MECH_ANALYTICS=false` on the staging env, redeploy, confirm reads route back to the subgraph
- [ ] Watch for `MechAnalyticsError` in the server logs during the first 24-48h of production traffic

## Rollback

Set `USE_MECH_ANALYTICS=false` in the deployment env. No revert needed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -84,5 +84,6 @@ MODE_RPC=__URL__
 SOLANA_RPC=__URL__
 
 # mech-analytics API (off-chain migration, server-only) — see docs/mech-analytics-migration.md
-MECH_ANALYTICS_URL=__URL__
+# The base URL is hardcoded in common-util/api/predict/mech-analytics.ts;
+# only the rollback lever is env-driven.
 USE_MECH_ANALYTICS= # unset = on; set to false (case-insensitive) to roll reads back to the subgraph
--- a/.env.example
+++ b/.env.example
@@ -85,4 +85,4 @@ SOLANA_RPC=__URL__
 
 # mech-analytics API (off-chain migration, server-only) — see docs/mech-analytics-migration.md
 MECH_ANALYTICS_URL=__URL__
-USE_MECH_ANALYTICS=true
+USE_MECH_ANALYTICS= # unset = on; set to false (case-insensitive) to roll reads back to the subgraph
--- a/.env.example
+++ b/.env.example
@@ -85,4 +85,4 @@ SOLANA_RPC=__URL__
 
 # mech-analytics API (off-chain migration, server-only) — see docs/mech-analytics-migration.md
 MECH_ANALYTICS_URL=__URL__
-USE_MECH_ANALYTICS=false
+USE_MECH_ANALYTICS=true
--- a/common-util/api/predict/mech-analytics.ts
+++ b/common-util/api/predict/mech-analytics.ts
@@ -7,8 +7,12 @@ import { QMR_MAX_AGE_DAYS } from 'common-util/constants';
  * Full details: docs/mech-analytics-migration.md
  */
 
-/** One switch that turns on all mech-analytics read paths. */
-export const USE_MECH_ANALYTICS = process.env.USE_MECH_ANALYTICS === 'true';
+/** One switch that turns on all mech-analytics read paths. Default ON
+ * after the mech-analytics cutover (undelivered rows admitted as shell
+ * shape in mech-analytics PR #36 + placeholder repair against the
+ * marketplace subgraph); explicit ``USE_MECH_ANALYTICS=false`` still
+ * routes reads back through the subgraph as a rollback lever. */
+export const USE_MECH_ANALYTICS = process.env.USE_MECH_ANALYTICS !== 'false';
 
 const PAGE_SIZE = 5000; // endpoint max limit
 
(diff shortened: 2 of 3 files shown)


common-util/api/predict/mech-analytics.ts
```diff
@@ -7,8 +7,12 @@ import { QMR_MAX_AGE_DAYS } from 'common-util/constants';
...
  */
 
-/** One switch that turns on all mech-analytics read paths. */
-export const USE_MECH_ANALYTICS = process.env.USE_MECH_ANALYTICS === 'true';
+/** One switch that turns on all mech-analytics read paths. Default ON
+ * after the mech-analytics cutover (undelivered rows admitted as shell
+ * shape in mech-analytics PR #36 + placeholder repair against the
+ * marketplace subgraph); explicit ``USE_MECH_ANALYTICS=false`` still
+ * routes reads back through the subgraph as a rollback lever. */
+export const USE_MECH_ANALYTICS = process.env.USE_MECH_ANALYTICS !== 'false';
```
Tanya-atatakai: `docs/mech-analytics-migration.md` still says "**Status:** `USE_MECH_ANALYTICS` is off everywhere" and "Default: off" — and this comment points readers to that doc. the doc header even asks to be updated when the flag changes :D can we update the status + default lines in this PR?

Tanya-atatakai: question: previews / local dev that never set `MECH_ANALYTICS_URL` used to silently take the subgraph path; now they hit "MECH_ANALYTICS_URL not set" on every refresh run by default (roi-distribution soft-fails and keeps stale data, tool-accuracy drops senders with a console.error). is that intended for non-prod envs, or should a missing URL route back to the subgraph? lmk if all envs already have the URL set

Tanya-atatakai: tiny, up to you: the rollback lever requires the exact string `false` — someone typing `FALSE`/`0` during an incident silently won't roll back and won't notice. wonder if we should accept a case-insensitive `false`? same for townhall-kpis #151 if yes

.env.example
```diff
@@ -85,4 +85,4 @@ SOLANA_RPC=__URL__
 
 # mech-analytics API (off-chain migration, server-only) — see docs/mech-analytics-migration.md
 MECH_ANALYTICS_URL=__URL__
-USE_MECH_ANALYTICS=false
+USE_MECH_ANALYTICS=true
```
Tanya-atatakai: tiny: `=true` is now a no-op (unset already means on). maybe document the rollback lever instead, wdyt:
```suggestion
USE_MECH_ANALYTICS= # unset = on; set to false to roll reads back to the subgraph
```

Tanya-atatakai (dismissed):
approving with caveats — the flip itself is right; please sync the migration doc with the new default (comment below), the rest is optional

common-util/api/predict/mech-analytics.ts
DIvyaNautiyal07: Good catch, fixed. Rewrote the status header, the flag/default sentence in the intro, and ticked off steps 1-3 of the rollout section. Step 4 (remove flag + subgraph paths) stays as the follow-up after the watch window.

DIvyaNautiyal07: Not sure prod always has it, so I added the guard defensively. USE_MECH_ANALYTICS now also checks !!process.env.MECH_ANALYTICS_URL, so an env without the URL falls back to the subgraph instead of hitting the fetch error. Previews and local dev get the safe default with no config change on their end.

DIvyaNautiyal07: Done, made it case-insensitive on both PRs. FALSE / False / faLSE all trigger the rollback now.

.env.example
DIvyaNautiyal07: Applied your suggestion, thanks.

OjusWiZard (dismissed):

OjusWiZard (approved):

Tanya-atatakai (approved):

